PR #13, opened 2015-05-14
Switch function definitions from mixed K&R to consistent ANSI C

stefanreinauer:
flex was using K&R function defintions for some functions and
ANSI C style in others, sometimes even in the same file. Change
the code to consistently use ANSI C.

Signed-off-by: Stefan Reinauer <stefan.reinauer@coreboot.org>


westes:
This is now in master. Thanks for the submission.

PR #16, opened 2015-11-05
Streamrefs

Mightyjo:
All,

I've updated the c++ portions of flex.skl by using references to clarify the ownership and lifetime of iostream objects passed to yyFlexLexer.  I retained the pointer-based interfaces as well to avoid breaking old code (at least until the next major release).  The test suite passes all 114 checks.  I checked that cxx_yywrap is working, too.  If there were other tests not run by 'make check' please let me know, especially if I broke them.

I've updated the docs to reflect my changes.  I also added notes to clarify that yyFlexLexer is declared in FlexLexer.h.  

I updated Makefile.am in src/ to let the maintainer-clean target remove skel.c because make's dependency checking doesn't always notice when it needs refreshing.  Similarly, I updated Makefile.am in tests/ to let the clean target remove all the files generated by flex and bison.  The changes in tests/ were necessary because the test sources don't depend on files in src/ at all.

Hope these small changes are useful.

-Joe

### DIFF
--- a/doc/flex.texi
+++ b/doc/flex.texi
@@ -3764,8 +3764,7 @@ defaults to generating the scanner to the file @file{lex.yy.cc} instead
 of @file{lex.yy.c}.  The generated scanner includes the header file
 @file{FlexLexer.h}, which defines the interface to two C++ classes.
 
-The first class,
-@code{FlexLexer},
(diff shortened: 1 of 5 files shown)


Mightyjo:
Closing this pull request to split out some build system commits to another branch.

PR #17, opened 2015-11-08
Dependency tracking updates to Makefile.am in src and tests

Mightyjo:
These are the isolated build system changes from pull request #16.

Added skel.c (as BUILT_SOURCES) to MAINTAINERCLEANFILES so 'make maintainer-clean' will remove it on demand, if needed.  Did this because make's dependency tracking doesn't always notice that flex.skl has changed.  I don't know why that's happening yet.  Also updated BUILT_SOURCES in tests to include several that weren't being cleaned up.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -67,6 +67,8 @@ EXTRA_DIST = \
 BUILT_SOURCES = \
 	skel.c
 
+MAINTAINERCLEANFILES = $(BUILT_SOURCES)
+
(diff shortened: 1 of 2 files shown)


westes:
I'm reviewing this and I've changes your reference to built_SOURCES to the literal skel.c in mainternercleanfiles as it is possible that not all built_SOURCES in flex's future will be things that should go in that variable.

westes:
I've accepted this with the changes noted above and I reworded your commit message. It's in master and is pushed to github.

westes:
b

PR #19, opened 2015-11-08
Make tests/*.l.c, etc. depend upon the built flex binary.

Mightyjo:
Added dependencies on the built flex binary to the test rules that use it so they'll rebuild whenever the binary has been updated.


westes:
This is now in master, thanks!

PR #20, opened 2015-11-18
Bootstrap the flex binary

Mightyjo:
Added a bootstrapping cycle to src/Makefile.am to ensure changes to scan.l get built into the new flex binary.  

This isn't perfect yet.  
* Dependencies on flex.skl still need to be added.  
* I might need to split 4053d8b into its own PR, though it is required for this one to pass distcheck.
* The "stage1" files aren't tracked by depcomp yet.


Mightyjo:
Looks like scan_stage1.c is getting into the dist tarball.  That is not what we want.

Mightyjo:
I wanted automake to track scan_stage1.c for me, but there's no way to do that without making the bootstrap a lot more complicated.  I think I'd have to duplicate the whole src/ tree at build time instead of just scan.l.

westes:
Yes, the change to ensure that test suites are built with the in-tree header should be a separate PR. That's a bug, regardless of how this ends up.

westes:
Please rebase this off of master. I pulled out the tests/Makefile.am changes and reworked them to be simpler. I then deleted my system's copy of FlexLexer.h and ran make distcheck successfully.

Mightyjo:
I will rebase off of master.

westes:
I've cut my own branch at https://github.com/westes/flex/tree/mightyjo-bootstrap and integrated the lastest changes to master into this branch. Please update and push to the tip of your bootstrap branch.

Mightyjo:
Brought my branch up to date with master.  Do I still need to rebase to remove 4053d8b?

westes:
Yes, the old CPPFLAGS are still present on this pr.

Mightyjo:
Okay, rebased to remove 4053d8b.  Cleaned out my tweaked version numbers from configure.ac and NEWS so I won't clobber them in master.  Did a 'git clean -xdf' and then passed 'make distcheck.'  I think we're good to go.

westes:
You're just missing this:

https://github.com/westes/flex/commit/22fb17e279621b49c487421c2ab5e24aa18d03d9

westes:
thanks!

Mightyjo:
> You're just missing this:
> 22fb17e

Hm, I thought I got that in 7fde0d3.  Looking into it.

westes:
oh you did. i posted without noticing your update.

It's all good now. There were a couple of changes on your branch that I manually put on master. One to configure.ac to put AC_INIT on one line and there was another CPPFLAGS that I missed on master somehow.

Mightyjo:
Oh, ha.  I thought I'd broken the AC_INIT line in two myself.

westes:
I s'pose we could ask git blame, But I know I saw the split when asking autoupdate a while back. In any case it's fixed on master...

westes:
I'm looking at this and I have a couple thoughts:

We want to build flex from itself. Specifically, flex should be built from the current version of flex and we also want the generated .l and .c files in flex to be built from the current version of flex. If we accomplish the former, we accomplish the latter. (The proof is left as an exercise for the reader.)

I think this implies a nodist_ prefix all the way down for the intermediate/stage1 build.

It does also imply that we need two almost identical programs built. In particular, we have to list all the sources because, say, we add a new command line option, then stage1flex should be rebuilt in that case.

I think your proposed change is almost correct. I'll play with it in the next few days and see what I can come up with.

Mightyjo:
Did you make any headway on the nodist_ question?

westes:
I'm cleaning up src/Makefile.am first. That's what the latest changes on master are about. I think I see how to do this. I wanted more clarity in what's in automake's court.

The plan is to introduce an entire nodist_bin_PROGRAM with a complete list of source files, splitting off the "common" files as you had done and only recompiling the sources that need that as flex bootstraps itself.

On Sunday, 29 November 2015,  6:13 pm -0800, Joseph Langley <notifications@github.com> wrote:

> Did you make any headway on the nodist_ question?
> 
> On Sat, Nov 21, 2015 at 7:17 PM, Will Estes <notifications@github.com>
> wrote:
> 
> > I'm looking at this and I have a couple thoughts:
> >
> > We want to build flex from itself. Specifically, flex should be built from
> > the current version of flex and we also want the generated .l and .c files
> > in flex to be built from the current version of flex. If we accomplish the
> > former, we accomplish the latter. (The proof is left as an exercise for the
> > reader.)
> >
> > I think this implies a nodist_ prefix all the way down for the
> > intermediate/stage1 build.
> >
> > It does also imply that we need two almost identical programs built. In
> > particular, we have to list all the sources because, say, we add a new
> > command line option, then stage1flex should be rebuilt in that case.
> >
> > I think your proposed change is almost correct. I'll play with it in the
> > next few days and see what I can come up with.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/westes/flex/pull/20#issuecomment-158693605>.
> >
> 
> 
> ---
> https://github.com/westes/flex/pull/20#issuecomment-160498169

-- 
Will Estes
westes575@gmail.com

westes:
Thanks for submitting this. Your solution wasn't quite right, as you noted, but you did prove that bootstrapping flex was pretty simple. See master for what I ended up doing.

I don't think we need to worry about changes in flex.skl / skel.c since skel.c is not generated by flex.

PR #21, opened 2015-11-28
Replace Char macro with 'unsigned char'

Mightyjo:
Thanks to Michael McConville for pointing out that the old Char macro causes problems with static analysis.  The macro has been removed and replaced with 'unsigned char' throughout the src.


westes:
Thanks for submitting this. It's in master now and will be included in the next release.

PR #22, opened 2015-12-02
Replace copy_string() with strdup()

mmcco:
Is there any reason not to use the stdlib alternative here? I can't imagine so.


westes:
See 399e94f. This is now in master and will be included in the next release of flex.

I changed your commit message because "git log".

Much appreciation for this sort of work.

PR #23, opened 2015-12-03
Enhance configure.ac checks for maintenance programs

Mightyjo:
Add test for presence of texi2dvi program. Give notice if texi2dvi
is unavailable and set TEXI2DVI=: to avoid giving users headaches.

Enhance tests for bison and help2man with notices when the programs
aren't found.  Set their program variables to use the missing script
in build-aux since it's compatible with them.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -48,8 +48,23 @@ AC_PROG_LN_S
 AC_PROG_AWK
 AC_PROG_INSTALL
 
-AC_PATH_PROG(BISON, bison,bison)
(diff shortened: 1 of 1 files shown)


westes:
Thanks for this. See https://github.com/westes/flex/commit/9e24fe5ea9ae7b2e2193fa3ad881416f12bf43cf for the details of what I pushed up. We prefer gtexi2dvi over texi2dvi and I normalized the commit message itself. Much appreciated!

PR #24, opened 2015-12-04
C++ cleanup

akimd:
Hi,

The following commits clean up the C++ code.

I have a number of failures of the test suite.  Unrelated to these patches, master fails.  Is this known?  It's mostly `flex_int32_t` errors: it is undefined.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -29,5 +29,5 @@
 #Makefile first.)
 
 touch ChangeLog
-libtoolize --install --force
+libtoolize --install --force || glibtoolize --install --force
 autoreconf --install --force
--- a/examples/testxxLexer.l
+++ b/examples/testxxLexer.l
@@ -39,13 +39,13 @@ number	{num1}|{num2}
 			}
 		}
 
-{number}	FLEX_STD cout << "number " << YYText() << '\n';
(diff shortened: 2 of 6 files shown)


westes:
> I have a number of failures of the test suite. Unrelated to these patches, master fails. Is this known? It's mostly flex_int32_t errors: it is undefined.

No, this is not known. I'm getting all tests passing.

Could you post to the mailing list about this?

akimd:
Well, it seems that the test suite is not robust to out-of-tree builds, and to concurrent execution.  Without -j and in-place it works properly.

westes:
I removed the removal of the c++ declaration. There is an edge case where we mix C and C++ compilation, so, at least for now, I'd like to leave that in. Other than that, I rewrote your commit messages to conform more to standard practice and to explain the reasoning better.

Thanks for the cleanup.

Regarding the test suite, could you open issues describing the two problems? They should be separate issues as they are two separate bugs. Building outside the tree is likely fixable by explicitly specifying srcdir and builddir as needed. I'm surprised by the -j failure, though. It likely means there's some dependency among the tests that I haven't noticed before.

Posting logs of the make check runs would help debug the problems. Likely:

make -k check


along with the other options you're passing would help so we can see the entire run -- such as it is.

PR #25, opened 2015-12-05
Replace copy_unsigned_string() with xstrdup().

mmcco:
Like copy_string(), copy_unsigned_string() is just a clone of the
stlib's strdup(). We only use it twice. I'm pretty confident that char
signedness is irrelevant in this case.


westes:
Thanks!

This is now in master. I'll review your other pr before pushing to origin.

PR #26, opened 2015-12-05
Use NULL rather than (type *) 0.

mmcco:


westes:
Thanks for this. It's now in master and pushed to origin along with the previous pr.

PR #27, opened 2015-12-08
Remove allocation casts

mmcco:
These casts were necessary decades ago, but no more.

There's a lot more to do, but this diff is big enough as-is.


westes:
Thanks for this cleanup. I had to fix one  parenthesis in one of the tests, but otherwise, lovely.

PR #29, opened 2015-12-09
Remove needless NULL-checks before free()

mmcco:


westes:
Thanks, it's on master.

PR #30, opened 2015-12-09
isascii is POSIX-defined

mmcco:
I'm pretty sure we can count on it being there.


westes:
Thanks. This is in master and will be pushed when I review your other pr.

PR #31, opened 2015-12-09
Remove alloca() configuration

mmcco:
alloca() is deprecated and dangerous. Unless I'm missing something, we
should never be using it.


westes:
Again, thanks. This is in master and pushed to origin, along with the previous pr.

PR #32, opened 2015-12-09
Remove a couple needless allocation casts

mmcco:


westes:
Thanks. This is on master. I'll push after looking at #33

PR #33, opened 2015-12-09
Replace zero_out() with the stdlib's memset

mmcco:
memset() is a stdlib function, so there's no need to recreate it.


westes:
Thanks. Also on master and pushed to origin, along with #32.

PR #34, opened 2015-12-12
Add reallocarray

mmcco:
This is taken from OpenSSH Portable, which in turn takes it from
OpenBSD.

reallocarray wraps the stdlib's realloc function. It takes two size
arguments and checks for overflow, like calloc, but doesn't zero the
memory. Therefore, it allows us to do overflow-safe array reallocations
and overflow-safe unzeroed array allocations, which the stdlib
allocation functions don't.

We have a bunch of needlessly specific array allocation macros, none of
which check for overflow. reallocarray should be able to replace them.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -125,7 +125,15 @@ AC_FUNC_ALLOCA
 AC_FUNC_FORK
 AC_FUNC_MALLOC
 AC_FUNC_REALLOC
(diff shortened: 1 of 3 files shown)


westes:
This is now in master, albeit with some changes on top of what you submitted. The upshot is that you can just use reallocarray now.

PR #35, opened 2015-12-12
Remove MS-DOS and non-POSIX-compliant-VMS macros

mmcco:
I'd venture a guess that no one builds on these platforms anymore.


westes:
Thanks for this. It's now on master and pushed to origin. It will be included in the next release of flex.

PR #36, opened 2015-12-12
Begin removing PROTO macro

mmcco:
I've read that all modern C and C++ compilers define __STDC__, so PROTO
is unneeded. In fact, we already have a bunch of prototypes that don't
use it. However, it's used in so many places and in so many formats that
we should probably remove it in multiple commits.


westes:
Thanks. This is now on master and pushed to origin.

PR #37, opened 2015-12-12
Unifdef SHORT_FILE_NAMES

Pyrohh:
It's never defined; see 13b5b214f53d1c3354a7ab910bd160c126df1331

While here also remove a stray MS_DOS ifdef.


westes:
Thanks. This is on master and pushed to origin.

PR #39, opened 2015-12-13
build/test fixes

vapier:


westes:
Thanks for this. It's now in master and will be pushed to origin after reviewing other outstanding pr's.

PR #40, opened 2015-12-13
Removed more instances of PROTO

mmcco:
And did some ANSIfication while at it.


westes:
Thanks. THis is now on master and pushed to github.

PR #41, opened 2015-12-13
Remove two superfluous, distracting macros

mmcco:
Simply incrementing or decrementing the static variable is much clearer
and leaves the programming with less to remember.


westes:
Thanks. This is now on master and will be pushed to origin after I review the other pr.

PR #42, opened 2015-12-13
README tweaks

mmcco:
Are you sure we want to keep taking bug reports on Sourceforge? It looks
like the GitHub repo's issue tracker is enabled, and I suspect that most
people's opinion is the less Sourceforge the better.

Also, a little trivial reformatting.

### DIFF
--- a/README
+++ b/README
@@ -8,11 +8,6 @@ flex can be found at:
 
 http://flex.sourceforge.net/
 
-Bug reports should be submitted using the SourceForge Bug Tracker for
-flex at:
-
-http://sourceforge.net/tracker/?group_id=97492&atid=618177
-
 The flex codebase is kept in git at:
 
 https://github.com/westes/flex
@@ -35,24 +30,15 @@ http://sourceforge.net/mail/?group_id=97492
 The flex distribution contains the following files which may be of interest:
 
 README - This file.
-
 NEWS - current version number and list of user-visible changes.
-
 INSTALL - basic installation information.
-
 ABOUT-NLS - description of internationalization support in flex.
-
 COPYING - flex's copyright and license.
-
 doc/ - user documentation.
-
 examples/ - containing examples of some possible flex scanners and a
-few other things. See the file examples/README for more details.
-
+	few other things. See the file examples/README for more details.
 TODO - outstanding bug reports, desired features, etc.
-
 tests/ - regression tests. See TESTS/README for details.
-
 po/ - internationalization support files.
 
 You need the following tools to build flex from the maintainer's


westes:
Thanks. I touched up a bit more beyond this, but yes, the move away from sourceforge does need to happen.

That being said, I'm using them for mailing lists and for file distribution, neither of which github does.

mmcco:
By file distribution, do you mean tarballs? If so, GitHub definitely does that. I'd say it's much better than Sourceforge for it, too, because it uses HTTPS, doesn't bundle malware, and seems to have way better opsec.

westes:
So with sourceforge, I can upload a tarball. I've not seen how to do that with github. Link me if it's a thing?

On Sunday, 13 December 2015,  5:05 pm -0800, Michael McConville <notifications@github.com> wrote:

> By file distribution, do you mean tarballs? If so, GitHub definitely does that. I'd say it's much better than Sourceforge for it, too, because it uses HTTPS, doesn't bundle malware, and seems to have way better opsec.
> 
> ---
> https://github.com/westes/flex/pull/42#issuecomment-164316988

-- 
Will Estes
Flex Project Maintainer
westes575@gmail.com
https://github.com/westes/flex

mmcco:
When you make a release, it automatically creates tarballs. Are you thinking of uploading binaries as well? [This announcement](https://github.com/blog/1547-release-your-software) seems to suggest that you can do that too.

mmcco:
Here's the guide, which describes how to include binaries: https://help.github.com/articles/creating-releases/

westes:
Ah, but github releases are just tars of the git tags. I need the output of "make dist" to be the release. So I'd be throwing away their tarball/zip in favor of another one.

On Sunday, 13 December 2015,  6:48 pm -0800, Michael McConville <notifications@github.com> wrote:

> Here's the guide, which describes how to include binaries: https://help.github.com/articles/creating-releases/
> 
> ---
> https://github.com/westes/flex/pull/42#issuecomment-164327249

-- 
Will Estes
westes575@gmail.com

mmcco:
You should be able to do that with the binary file addition.

westes:
Yes, I see how that's possible. I'll start trying this out. I'll backfill the 2.6.0 release with tarballs as a trial.

On Monday, 14 December 2015, 11:35 am -0800, Michael McConville <notifications@github.com> wrote:

> You should be able to do that with the binary file addition.
> 
> ---
> https://github.com/westes/flex/pull/42#issuecomment-164536486

-- 
Will Estes
Flex Project Maintainer
westes575@gmail.com
https://github.com/westes/flex

PR #43, opened 2015-12-15
Remove prototype for main()

Pyrohh:
It's not called anywhere else so the prototype is not needed.
See the C99 standard [1], section 5.1.2.2.1 for more info.

[1]: http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf


westes:
Thanks. THis is in master and will be included in the next release of flex.

PR #44, opened 2015-12-15
Remove more instances of PROTO

Pyrohh:
This PR is kept small for ease of review; I'll remove the remaining PROTO instances in a follow up PR.


westes:
Thanks. This is now in master and will be included in the next release.

PR #45, opened 2015-12-21
buf.c: use snprintf

sparkhom:
This changes the sprintf function call to snprintf in buf.c, which is a safer string handling function. This is a downstream change from OpenBSD's version of flex.


westes:
Thanks for this. It's on master and will be included in the next relase of flex.

PR #46, opened 2015-12-25
Remove remaining use of PROTO

Pyrohh:
Follow up to #44.


westes:
Thanks. This is on master and will be pushed shortly.

PR #47, opened 2015-12-25
Cleanup __STDC__ #ifdefs

Pyrohh:
Assuming a compiler conforming to the ISO C standard is used, i.e.,
__STDC__ is defined to 1, the ifdefs surrounding YY_USE_CONST can be
greatly simplified.


westes:
Thanks. This is on master and will be pushed shortly.

PR #48, opened 2015-12-25
Replace basename2() with basename(3)

Pyrohh:
Please see the commit messages for more information.

Note that the first commit is done primarily to ease review.


westes:
Thanks for this. It's in master and is pushed to origin.

PR #50, opened 2016-01-27
Fix declaration mismatch in yy_fatal_error

tklauser:
The prototype declares yy_fatal_error parameter as "const char msg[]"
while the definition uses "const char* msg" (introduced by commit
e9d5fc713f61b) which causes the sparse static checkers to produce an
error.

Fix this by adjusting the definition to use "const char* msg" as well.
Also change the C++ version accordingly so it matches the declaration in
FlexLexer.hpp.


westes:

PR #51, opened 2016-01-29
Sparse fixes

tklauser:
Two more changes to fix warnings/errors emitted by the sparse static checker on the generated lexer.


westes:
Thanks. This is on master now and will be included in the next release of flex.

PR #52, opened 2016-02-16
Converted K&R style function definitions to ANSI C style

tklauser:
Consistently make use of the ANSI C function definition style instead of
the K&R style.


westes:
Thanks. This is in master and will be included in the next release of flex.

Issue #56, opened 2016-02-17
%option noline generates and error message


srivasta:
Hi,

  This bug was reported by a Debian user in 
  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=729693

[credit.l.txt](https://github.com/westes/flex/files/135028/credit.l.txt)

    > flex credit.l
    credit.l:2: unrecognized %option: line
    > egrep option credit.l 
    %option noyywrap
    %option noline

Manoj

westes:
As per the most recent change to master, this is now resolved.

Issue #55, opened 2016-02-17
flex --noline outputs a #line directive


srivasta:
Hi,

  This issue was reported by a Debian user in [Bug 729927]( https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=729927)

The command
   flex --noline name.lex 

generates the line
   #line 3 "lex.yy.c"

This causes some confusion when:
1: line 3 is really line 2, and,
2: when (in order to compile using C++) the file has
   been renamed to (something like) lex.yy.cpp causing
   error messages to specify file lex.yy.c instead of
   lex.yy.cpp (or some other name).

   Manoj

westes:
As per recent changes to master, this is now resolved.

PR #59, opened 2016-02-23
Line directive fixes

tklauser:
Two fixes related to using the #line directives (#55 and #56 as reported by @srivasta).


westes:
Thanks for this. It's in master and pushed to origin. It will be included in the next release of flex.

PR #65, opened 2016-03-14
Fixes yyl compare with unsigned warning

bkchr:


westes:

PR #69, opened 2016-03-31
Fix potential buffer overflow in strncat()

tklauser:
When using clang/llvm 3.8 to compile flex, the following warning is
emitted:

main.c:378:27: warning: the value of the size argument in 'strncat' is too large, might lead to a buffer overflow [-Wstrncat-size]
                                        strncat(m4_path, m4, sizeof(m4_path));
                                                             ^~~~~~~~~~~~~~~
main.c:378:27: note: change the argument to be the free space in the destination buffer minus the terminating null byte
                                        strncat(m4_path, m4, sizeof(m4_path));
                                                             ^~~~~~~~~~~~~~~
                                                             sizeof(m4_path) - strlen(m4_path) - 1

Fix it up by using the solution proposed by the warning message.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -375,7 +375,7 @@ void check_options (void)
 					strncpy(m4_path, path, sizeof(m4_path));
 					m4_path[endOfDir-path] = '/';
 					m4_path[endOfDir-path+1] = '\0';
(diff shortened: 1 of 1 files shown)


westes:
Thanks. This is now pushed to master

PR #74, opened 2016-05-23
Delete action_m4_define from misc.c

drbo:
This function was not used and always triggered a fatal error when run.


westes:
Thanks! THis is on master and will be included in the next release of flex.

PR #81, opened 2016-09-04
Fix escaping of `[[` and `]]` in strings

DemiMarie:
Previously, `[[` and `]]` were not escaped in strings, which led
to bad interactions with m4.

Also, don't break strings on newline, as GCC et al support whitespace
between a backslash and the subsequent newline.


westes:
Thanks for this. It's merged to master and will be in the next release of flex.

PR #83, opened 2016-09-23
Remove support for K&R C

DemiMarie:
This removes support for generating K&R C scanners.  C89 was at least 26 years ago, and every compiler I know of supports at least C89.


westes:
Thanks for this; it's now on master and will be included in the next relase. I did rebase it down to one commit since your other two commits were fixups of the main one.

PR #84, opened 2016-09-25
Simplify some shell code

DemiMarie:
This simplifies some shell code, and makes the sed replacements during skeleton generation reversible.


westes:

PR #85, opened 2016-09-25
Fix m4 quotation

DemiMarie:
This fixes almost all of the M4 quoting problems, except for those relating to section 3 of the output file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,8 @@ libtool
 stamp-*
 m4/
 build-aux/
+build/
+*.po
+*.o
+*.so
+*.a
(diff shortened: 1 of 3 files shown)


westes:
The changes to .gitignore are incorrect: *.po files are explicitly managed by git as they're created by the translation project. *.{a,o,so} already are ignored, I think by some default mechanism. The directory build/ is not something that flex creates, so if you want to ignore it, you should put that in your own personal ~/.gitignore or similar.

westes:
w

westes:
I've put this onto master, minus the .gitignore stuff. That will be in the next release of flex. Thanks for the fixes.

PR #86, opened 2016-09-27
Fix M4 quotation in section 3

DemiMarie:
This patch fixes M4 quotation in section 3, using the same technique used elsewhere.  This uncovered numerous bugs in the test suite due to using `%option prefix="test"` but not prefixing all identifiers used by Flex correctly.


westes:
Would you add documentation of the new options to the manual?

DemiMarie:
I left it deliberately undocumented, as it is only meaningful because of an
implementation detail of Flex.  It really is for use in Flex's test suite
and nowhere else – in fact, I would set it at compile time if possible.

westes:
Thanks. this is on master and will be included in the next release of flex.

PR #87, opened 2016-10-04
Fix yyrestart(NULL) SEGV.

Mightyjo:
When I made the scanner use C++ stream references, I screwed up the semantics of yyrestart(NULL). 

If I ever get finished with my C++ overhaul this will cease being a problem, but for now I've added a test and fixed my segfault.  Sorry, world!


westes:
Thanks for this. It's on master and will be in the next release of flex.

PR #90, opened 2016-10-14
Improved M4 quotation

DemiMarie:
This fixes M4 quotation of certain strings beginning with `yy` (in
section 3 of the input file only) and character literals.  The new
quotation method is also less brittle and faster.

Tests that relied on the old behavior were fixed.

Also, `yyconst` is no longer defined; use `const` (which it
unconditionally was defined to) instead.

### DIFF
--- a/src/dfa.c
+++ b/src/dfa.c
@@ -531,12 +531,12 @@ void ntod (void)
 		 */
 		if (gentables)
 			out_str_dec
-				("static yyconst %s yy_nxt[][%d] =\n    {\n",
+				("static const %s yy_nxt[][%d] =\n    {\n",
 				 long_align ? "flex_int32_t" : "flex_int16_t",
 				 num_full_table_rows);
 		else {
 			out_dec ("#undef YY_NXT_LOLEN\n#define YY_NXT_LOLEN (%d)\n", num_full_table_rows);
-			out_str ("static yyconst %s *yy_nxt =0;\n",
+			out_str ("static const %s *yy_nxt =0;\n",
 				 long_align ? "flex_int32_t" : "flex_int16_t");
 		}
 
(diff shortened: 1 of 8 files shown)


westes:

westes:
I'm having some issues around this which I didn't see earlier. Reopening to investigate.

westes:
This is what I'm seeing:

gcc -DHAVE_CONFIG_H -I.  -DLOCALEDIR=\"/usr/local/share/locale\" -I../intl   -g -O2 -MT stage1flex-scan.o -MD -MP -MF .deps/stage1flex-scan.Tpo -c -o stage1flex-scan.o `test -f 'scan.c' || echo './'`scan.c
scan.l: In function ‘flexscan’:
scan.l:918:33: error: expected ‘)’ before ‘;’ token
      add_action( "\tYY_BREAK]]\n" );
                                 ^
scan.l:922:5: error: expected ‘;’ before ‘}’ token
     }
     ^
scan.c: At top level:
scan.c:4337:1: error: expected identifier or ‘(’ before ‘]’ token
 ]])
 ^
scan.c:350:12: warning: ‘yy_get_next_buffer’ used but never defined
 static int yy_get_next_buffer (void );
            ^
Makefile:1077: recipe for target 'stage1flex-scan.o' failed

Pushed to https://github.com/westes/flex/tree/DemiMarie-different-quoting-pr so you can see exactly what I'm building against.

DemiMarie:
I think it has something to do with us using different bootstrap versions of Flex: I am using a version I built from Git rather recently, while you may be using a version that predates the M4 quoting fixes I have added.  That would explain the behavior you are seeing.

westes:
That could be it. I will install from git just before this change, then.

DemiMarie:
I just pushed a commit that should fix it.

westes:
Yes, that helped too. And so did installing from a more recent version. It's on master and will go in 2.6.2.

PR #95, opened 2016-10-22
Update flex.texi: Removed repedative wording

LaikaN57:
"Flex used to" was written twice at the beginning of Chapter 9, paragraph 2.


westes:
Thanks! It's on master and will be included in the next release. The diff is also pushed to the html copy on the manual.

PR #96, opened 2016-10-23
Fix M4 quotation in section 2 prologue and refactor duplicated code

DemiMarie:
This fixes M4 quotation in the section 2 prologue and refactors some duplicated code.  I believe that this is the last M4 quotation-related bug, except for the lack of M4 quotation in `#line` directives (but am not certain).

No tests yet.


DemiMarie:
Test added.

westes:

PR #101, opened 2016-10-25
doc: README.md formatting fixes

Explorer09:
* Wrap everything in the raw document in 72 char per line limit.
* Proper casing for terms "Git" and "GitHub" (don't try to look lazy).
* Add unordered list marks when needed.
* Say `configure && make && make install` and quoted for fixed-width
  font.


westes:
I think quoting the autogen.sh line would be better.

westes:
Thanks. It's on master and will be included in the next release of flex.

PR #109, opened 2016-10-26
Fix cross compiling

### LINKED ISSUE #78: flex 2.6.1 does not cross compile
As of 53e3088, flex runs stage1flex to bootstrap itself. Unfortunately, this fails when cross compiling, as stage1flex is built for the target architecture (not for the host machine).

I'm unsure how to fix this as my knowledge of GNU autotools is limited; but http://lists.gnu.org/archive/html/automake/2008-12/msg00016.html seems to suggest a way to make this work.

hobbitalastair:
Detect if the build lexer is sufficient. If not, build stage1flex but
warn if we are cross compiling as we may not be able to run the result.

This probably needs additional testing. I attempted to build a stage1flex using the build tools, however the changes became quite invasive as libcompat.la also needed to be rebuilt to use the build compiler and flags. This method seemed easier and is probably more acceptable for cross compiling.

Fixes #78.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -101,6 +101,16 @@ else
    AC_MSG_WARN(no indent program found: make indent target will not function)
 fi
 
+# Check for a modern flex.
+
+AC_MSG_CHECKING(if $LEX is $PACKAGE $PACKAGE_VERSION)
+AM_CONDITIONAL([STAGE1FLEX], [test "`$LEX --version`" != "$PACKAGE $PACKAGE_VERSION"])
+AM_COND_IF([STAGE1FLEX],
+           [AC_MSG_RESULT(no); \
+            test "x$build" != "x$host" && \
+            AC_MSG_WARN(cross compiling but using stage1flex)],
+           [AC_MSG_RESULT(yes)])
+
 # checks for headers
 
 AC_CHECK_HEADERS([inttypes.h libintl.h limits.h locale.h malloc.h netinet/in.h regex.h unistd.h])
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -6,16 +6,19 @@ LIBS = @LIBS@
 m4 = @M4@
 
 bin_PROGRAMS = flex
-noinst_PROGRAMS = stage1flex
 lib_LTLIBRARIES = \
 	libfl.la \
 	libfl_pic.la
+
+if STAGE1FLEX
+noinst_PROGRAMS = stage1flex
 
 stage1flex_SOURCES = \
 	scan.l \
 	$(COMMON_SOURCES)
 
 stage1flex_CFLAGS = $(AM_CFLAGS)
+endif
 
 flex_SOURCES = \
 	$(COMMON_SOURCES)
@@ -89,15 +92,21 @@ skel.c: flex.skl mkskel.sh flexint.h tables_shared.h tables_shared.c
 	  $(SHELL) $(srcdir)/mkskel.sh > $@.tmp
 	mv $@.tmp $@
 
+if STAGE1FLEX
 stage1scan.c: scan.l stage1flex$(EXEEXT)
 	./stage1flex$(EXEEXT) -o $@ $<
+stage1scan.c : parse.h
+else
+stage1scan.c: scan.l
+	$(LEX) -o $@ $<
+stage1scan.c : parse.h
+endif
 
 # make needs to be told to make parse.h so that parallelized runs will
 # not fail.
 
 main.c : parse.h
 scan.c : parse.h
-stage1scan.c : parse.h
 yylex.c : parse.h
 
 # Run GNU indent on sources. Don't run this unless all the sources compile cleanly.


westes:
I've had my eye on [AX_PROG_FLEX](https://www.gnu.org/software/autoconf-archive/ax_prog_flex.html#ax_prog_flex) and [AX_PROG_FLEX_VERSION](https://www.gnu.org/software/autoconf-archive/ax_prog_flex_version.html) to handle the autoconf piece of this. I'll look more closely at your method though. It's overall simple which is very good.

Explorer09:
Hey. May I ask some questions here? What is the purpose of the stage1flex program? Is it used in any place other than making stage1scan.c?

My point is that if this stage1flex is used only in a bootstrap step to generate the .c file, then this should   not be cross-compiled (but be compiled with $CC_FOR_BUILD, a.k.a. the build system's native compiler).

(And by the way, if there would be a temporary flex program built with the native compiler, then ideally the flex.1 man page should also depend on this instead of the final flex program (which might be cross-compiled and cannot be run directly).)

westes:
@Explorer09:

> What is the purpose of the stage1flex program?

Flex should be built with itself. Hence the bootstrap. That's all it's there for, but hence the cross compilation issues.

> My point is that if this stage1flex is used only in a bootstrap step to generate the .c file, then this should not be cross-compiled (but be compiled with $CC_FOR_BUILD, a.k.a. the build system's native compiler).

Yes that's right, although documentation on cross compilation is pretty sketchy, unfortunately. If you know of something more detailed, that'd be helpful.

> (And by the way, if there would be a temporary flex program built with the native compiler, then ideally the flex.1 man page should also depend on this instead of the final flex program

You're right; good catch.

Explorer09:
Well I'm not experienced in autotools, actually. But I have tried building a cross toolchain and for GCC I know at least there is $CC_FOR_BUILD that they use with a purpose. Maybe this [AX_PROG_CC_FOR_BUILD](https://www.gnu.org/software/autoconf-archive/ax_prog_cc_for_build.html) could be what we need.

Ok, so I get the brief idea: the `scan.l` is first made into `scan.c` which will be bundled in release tarball. Then, `stage1flex` will be built with this `scan.c` and other source files. Then `stage1flex` will compile `scan.l` and generate `stage1scan.c`. (Until now this is stage1flex's only purpose.) The final flex binary will be built with `stage1scan.c` and other source files.

A diagram like this...am I correct?
diagram1

westes:
@Explorer09 that is exactly what is happening.

hobbitalastair:
@Explorer09 I got as far as this patch using that approach.
The patch below does not actually build, because the libcompatbuild stuff is (still) not built for the correct architecture; it seemed much simpler to let the maintainer of the toolchain build a more modern host flex first if it was required (what I am doing currently...).

```
From 72efcb306297fbc5b9a2d4b34f7457810d922618 Mon Sep 17 00:00:00 2001
From: Alastair Hughes <hobbitalastair@gmail.com>
Date: Wed, 26 Oct 2016 11:39:39 +1300
Subject: [PATCH] Make flex cross compile

Build stage1flex using the build compiler, instead of cross compiling
it and then trying to run it.
---
 configure.ac               |   4 ++
 lib/Makefile.am            |  12 +++++
 m4/ax_prog_cc_for_build.m4 | 125 +++++++++++++++++++++++++++++++++++++++++++++
 src/Makefile.am            |  12 ++++-
 4 files changed, 152 insertions(+), 1 deletion(-)
 create mode 100644 m4/ax_prog_cc_for_build.m4

diff --git a/configure.ac b/configure.ac
index 666b69f..601abf9 100644
--- a/configure.ac
+++ b/configure.ac
@@ -42,6 +42,7 @@ AM_GNU_GETTEXT_VERSION([0.19])
 AC_PROG_YACC
 AM_PROG_LEX
 AC_PROG_CC
+AC_PROG_CC_FOR_BUILD
 AC_PROG_CXX
 AM_PROG_CC_C_O
 AC_PROG_LN_S
@@ -157,4 +158,7 @@ tools/Makefile
 tests/Makefile
 )
 
+# Test for cross compiling.
+AM_CONDITIONAL([CROSS], [test x$build != x$host])
+
 AC_OUTPUT
diff --git a/lib/Makefile.am b/lib/Makefile.am
index 011163e..ce2ce35 100644
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -2,3 +2,15 @@ noinst_LTLIBRARIES = libcompat.la
 libcompat_la_SOURCES = lib.c
 libcompat_la_LIBADD = $(LTLIBOBJS)
 
+if CROSS
+noinst_LTLIBRARIES += libcompatbuild.la
+libcompatbuild_la_SOURCES = lib.c
+libcompatbuild_la_LIBADD = $(LTLIBOBJS)
+
+libcompatbuild_LINK = $(LIBTOOL) --tag=CC --mode=link $(CC_FOR_BUILD) \
+		  $(CFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@
+
+$(libcompatbuild_OBJECTS): CC=$(CC_FOR_BUILD)
+$(libcompatbuild_OBJECTS): CFLAGS=$(CFLAGS_FOR_BUILD)
+$(libcompatbuild_OBJECTS): CPPFLAGS=$(CPPFLAGS_FOR_BUILD)
+endif
diff --git a/m4/ax_prog_cc_for_build.m4 b/m4/ax_prog_cc_for_build.m4
new file mode 100644
index 0000000..77fd346
--- /dev/null
+++ b/m4/ax_prog_cc_for_build.m4
@@ -0,0 +1,125 @@
+# ===========================================================================
+#   http://www.gnu.org/software/autoconf-archive/ax_prog_cc_for_build.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_PROG_CC_FOR_BUILD
+#
+# DESCRIPTION
+#
+#   This macro searches for a C compiler that generates native executables,
+#   that is a C compiler that surely is not a cross-compiler. This can be
+#   useful if you have to generate source code at compile-time like for
+#   example GCC does.
+#
+#   The macro sets the CC_FOR_BUILD and CPP_FOR_BUILD macros to anything
+#   needed to compile or link (CC_FOR_BUILD) and preprocess (CPP_FOR_BUILD).
+#   The value of these variables can be overridden by the user by specifying
+#   a compiler with an environment variable (like you do for standard CC).
+#
+#   It also sets BUILD_EXEEXT and BUILD_OBJEXT to the executable and object
+#   file extensions for the build platform, and GCC_FOR_BUILD to `yes' if
+#   the compiler we found is GCC. All these variables but GCC_FOR_BUILD are
+#   substituted in the Makefile.
+#
+# LICENSE
+#
+#   Copyright (c) 2008 Paolo Bonzini <bonzini@gnu.org>
+#
+#   Copying and distribution of this file, with or without modification, are
+#   permitted in any medium without royalty provided the copyright notice
+#   and this notice are preserved. This file is offered as-is, without any
+#   warranty.
+
+#serial 8
+
+AU_ALIAS([AC_PROG_CC_FOR_BUILD], [AX_PROG_CC_FOR_BUILD])
+AC_DEFUN([AX_PROG_CC_FOR_BUILD], [dnl
+AC_REQUIRE([AC_PROG_CC])dnl
+AC_REQUIRE([AC_PROG_CPP])dnl
+AC_REQUIRE([AC_EXEEXT])dnl
+AC_REQUIRE([AC_CANONICAL_HOST])dnl
+
+dnl Use the standard macros, but make them use other variable names
+dnl
+pushdef([ac_cv_prog_CPP], ac_cv_build_prog_CPP)dnl
+pushdef([ac_cv_prog_gcc], ac_cv_build_prog_gcc)dnl
+pushdef([ac_cv_prog_cc_works], ac_cv_build_prog_cc_works)dnl
+pushdef([ac_cv_prog_cc_cross], ac_cv_build_prog_cc_cross)dnl
+pushdef([ac_cv_prog_cc_g], ac_cv_build_prog_cc_g)dnl
+pushdef([ac_cv_exeext], ac_cv_build_exeext)dnl
+pushdef([ac_cv_objext], ac_cv_build_objext)dnl
+pushdef([ac_exeext], ac_build_exeext)dnl
+pushdef([ac_objext], ac_build_objext)dnl
+pushdef([CC], CC_FOR_BUILD)dnl
+pushdef([CPP], CPP_FOR_BUILD)dnl
+pushdef([CFLAGS], CFLAGS_FOR_BUILD)dnl
+pushdef([CPPFLAGS], CPPFLAGS_FOR_BUILD)dnl
+pushdef([LDFLAGS], LDFLAGS_FOR_BUILD)dnl
+pushdef([host], build)dnl
+pushdef([host_alias], build_alias)dnl
+pushdef([host_cpu], build_cpu)dnl
+pushdef([host_vendor], build_vendor)dnl
+pushdef([host_os], build_os)dnl
+pushdef([ac_cv_host], ac_cv_build)dnl
+pushdef([ac_cv_host_alias], ac_cv_build_alias)dnl
+pushdef([ac_cv_host_cpu], ac_cv_build_cpu)dnl
+pushdef([ac_cv_host_vendor], ac_cv_build_vendor)dnl
+pushdef([ac_cv_host_os], ac_cv_build_os)dnl
+pushdef([ac_cpp], ac_build_cpp)dnl
+pushdef([ac_compile], ac_build_compile)dnl
+pushdef([ac_link], ac_build_link)dnl
+
+save_cross_compiling=$cross_compiling
+save_ac_tool_prefix=$ac_tool_prefix
+cross_compiling=no
+ac_tool_prefix=
+
+AC_PROG_CC
+AC_PROG_CPP
+AC_EXEEXT
+
+ac_tool_prefix=$save_ac_tool_prefix
+cross_compiling=$save_cross_compiling
+
+dnl Restore the old definitions
+dnl
+popdef([ac_link])dnl
+popdef([ac_compile])dnl
+popdef([ac_cpp])dnl
+popdef([ac_cv_host_os])dnl
+popdef([ac_cv_host_vendor])dnl
+popdef([ac_cv_host_cpu])dnl
+popdef([ac_cv_host_alias])dnl
+popdef([ac_cv_host])dnl
+popdef([host_os])dnl
+popdef([host_vendor])dnl
+popdef([host_cpu])dnl
+popdef([host_alias])dnl
+popdef([host])dnl
+popdef([LDFLAGS])dnl
+popdef([CPPFLAGS])dnl
+popdef([CFLAGS])dnl
+popdef([CPP])dnl
+popdef([CC])dnl
+popdef([ac_objext])dnl
+popdef([ac_exeext])dnl
+popdef([ac_cv_objext])dnl
+popdef([ac_cv_exeext])dnl
+popdef([ac_cv_prog_cc_g])dnl
+popdef([ac_cv_prog_cc_cross])dnl
+popdef([ac_cv_prog_cc_works])dnl
+popdef([ac_cv_prog_gcc])dnl
+popdef([ac_cv_prog_CPP])dnl
+
+dnl Finally, set Makefile variables
+dnl
+BUILD_EXEEXT=$ac_build_exeext
+BUILD_OBJEXT=$ac_build_objext
+AC_SUBST(BUILD_EXEEXT)dnl
+AC_SUBST(BUILD_OBJEXT)dnl
+AC_SUBST([CFLAGS_FOR_BUILD])dnl
+AC_SUBST([CPPFLAGS_FOR_BUILD])dnl
+AC_SUBST([LDFLAGS_FOR_BUILD])dnl
+])
diff --git a/src/Makefile.am b/src/Makefile.am
index 77ddf4b..7221137 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -15,7 +15,17 @@ stage1flex_SOURCES = \
 	scan.l \
 	$(COMMON_SOURCES)
 
+if CROSS
+stage1flex_LINK = $(LIBTOOL) --tag=CC --mode=link $(CC_FOR_BUILD) \
+		  $(CFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@
+stage1flex_LDADD = ../lib/libcompatbuild.la @LIBINTL@
+
+$(stage1flex_OBJECTS): CC=$(CC_FOR_BUILD)
+$(stage1flex_OBJECTS): CFLAGS=$(CFLAGS_FOR_BUILD)
+$(stage1flex_OBJECTS): CPPFLAGS=$(CPPFLAGS_FOR_BUILD)
+else
 stage1flex_CFLAGS = $(AM_CFLAGS)
+endif
 
 flex_SOURCES = \
 	$(COMMON_SOURCES)
@@ -89,7 +99,7 @@ skel.c: flex.skl mkskel.sh flexint.h tables_shared.h tables_shared.c
 	  $(SHELL) $(srcdir)/mkskel.sh > $@.tmp
 	mv $@.tmp $@
 
-stage1scan.c: scan.l stage1flex$(EXEEXT)
+stage1scan.c: scan.l stage1flex$(BUILD_EXEEXT)
 	./stage1flex$(EXEEXT) -o $@ $<
 
 # make needs to be told to make parse.h so that parallelized runs will
-- 
2.10.1

```

Explorer09:
@hobbitalastair 
Thanks for your effort. However, when I deeply think of this issue I believe using $CC_FOR_BUILD might not be the best approach for us. If all that are needed to bootstrap flex are `scan.c` generated from `scan.l`, and `parse.c` from `parse.y`, I think the best way is to have both files prepared in the release tarball, _and_ write instructions on how to generate them for people who checked-out code from git.

Otherwise, I found it rather annoying when `scan.c` from release tarball and `stage1scan.c` are not bit-identical, and differ only in `#line` directives that include file names:

```
--- scan.c	2016-10-25 06:33:47.000000000 +0800
+++ stage1scan.c	2016-10-30 19:14:06.134995996 +0800
@@ -1,5 +1,6 @@
+#line 1 "stage1scan.c"
 
-#line 2 "scan.c"
+#line 3 "stage1scan.c"
 
 #define  YY_INT_ALIGNED short int
 
@@ -1982,9 +1983,9 @@
     add_action(M4QEND); \
      if (!indented_code) line_directive_out(NULL, 0);\
       } while (0)
-#line 1985 "scan.c"
+#line 1986 "stage1scan.c"
 
-#line 1987 "scan.c"
+#line 1988 "stage1scan.c"
 
 #define INITIAL 0
 #define SECT2 1
@@ -2250,7 +2251,7 @@
 	char nmdef[MAXLINE];
 
 
-#line 2253 "scan.c"
+#line 2254 "stage1scan.c"
 
 	while ( /*CONSTCOND*/1 )		/* loops until end-of-file is reached */
 		{
@@ -4153,7 +4154,7 @@
 #line 1014 "scan.l"
 YY_FATAL_ERROR( "flex scanner jammed" );
 	YY_BREAK
-#line 4156 "scan.c"
+#line 4157 "stage1scan.c"
 case YY_STATE_EOF(INITIAL):
 case YY_STATE_EOF(SECT2):
 case YY_STATE_EOF(CODEBLOCK):
```

Here, why bother making the stage1flex program and stage1scan.c when the scan.c from release tarball is already "perfect" to build the final flex?

Ultimately, this bootstrapping problem **only matters for people who checked-out from git repo**, right? So, what about writing bootstrapping instructions in autogen.sh?

Might be like this: (just a concept; not tested)

```
libtoolize --install --force || glibtoolize --install --force
autoreconf --install --force
# Bootstrap.
# For this stage scan.c and parse.c are generated using host's lex and yacc
./configure && make all doc || { echo "Can't bootstrap">&2 ; exit 1}
# Now let our flex generate its own scan.c
rm -f src/scan.c && make LEX="src/flex" src/scan.c ||
{ echo "Can't generate scan.c from flex just compiled">&2 ; exit 1 }
# Clean up. (Shouldn't delete scan.c, parse.c, or doc/flex.1)
make distclean
```

westes:
We need a way to generate the flex scanner from the identical version of flex. THat's the rub. Having the release tar ball is not always sufficient. I'll dig on this some and see what we can do.

Explorer09:
I wonder in which cases are the pre-generated scan.c not sufficient. It's not in my build cases, at least, nor in typical case of cross-building.

rossburton:
Just hit this trying to build flex from git in Yocto, which cross compiles.

AX_PROG_CC_FOR_BUILD is definitely the way forward.  Be aware that per-target CC requires a fairly new automake version, if that is a concern.  The traditional workaround is to build host binaries in a subdirectory, pulling sources from the parent.

Explorer09:
@rossburton Hello. Do you build Bison too in your use case / system?
I'm just wanting to know how people cross-compiles flex and whether this additional bootstrap procedure ("stage1flex") is really worth it.

For me, I would build flex and Bison together for the same toolchain. Ideally for my case every .l file and .y file should be generated by my flex and Bison versions to be built. This is what I personally want for a "bootstrap" procedure, not just generating .l files.

rossburton:
Yes, currently with flex 2.6 we build both a native bison and native flex
2.6 (as that release needed an existing and matching flex binary if it
couldn't build one itself). This isn't really a problem for us, it's the
changes since 2.6 which break in cross-compilation.

Ross

Explorer09:
@rossburton
Precisely it's 2.6.1 that breaks cross-compilation. 2.6.0 does not have this problem.

westes:
@rossburton, no, we already assume / require recent automake. I do stick to already-released versions since bootstrapping flex is kind of a pain, depending on what your setup/need is, though.

I've put out a branch https://github.com/westes/flex/tree/cross taking the patch listed in the comments on this pr and touching it up a fair bit.

Tell me what you all think?

westes:
Would you resubmit this given the latest round of changes to the build system? In particular, libcompat is no longer a thing.

hobbitalastair:
I've opened #153 which uses a modified version of the approach listed in the comments and discussed at c74acb7c5d1. Thanks for continuing to look at this!

hobbitalastair:
Closing - I believe that the approach in #153 is preferred.

PR #110, opened 2016-10-27
Only regenerate the man page when required

### LINKED ISSUE #108: Build always regenerating manpage
Since <strike>84ca965</strike>, if <code>help2man</code> is not installed configure will warn but <code>make install</code> will fail <strike>since <code>flex.1</code> is never actually generated</strike>.

Ideally either the build would succeed without installing the man page or configure would fail if <code>help2man</code> was not installed.

EDIT: I seem to be completely wrong here. I'll take another look at what was causing the failure.

hobbitalastair:
Make the flex binary an order-only prerequisite, and add back the prerequisites from before 7cfb440. This prevents rebuilding the man page whenever the flex binary is rebuilt, which causes problems if <code>help2man</code> is not installed and will never work when cross compiling.

Fixes #108.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -23,8 +23,8 @@ CLEANFILES = \
 	flex.vr \
 	flex.vrs
 
-flex.1: ../src/flex$(EXEEXT)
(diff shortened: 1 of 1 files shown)


westes:
the built flex binary is in "$(top_builddir)/src/, not necessarily ../src.

Also, with multiple dependencies, $< needs to be changed to...something else.

westes:
I touched this up, added a NEWS entry and put it on master. Thanks for your submission!

PR #111, opened 2016-10-27
build: fix false negatives for help2man and texi2dvi

hobbitalastair:
<code>HELP2MAN</code> and <code>TEXI2DVI</code> (or the corresponding ac_prog variables) will never be zero length as they fall back to the missing script; check for the fall back and warn on that instead of *always* warning.

I've also made the ordering for both warnings consistent.


westes:
Thanks! I slightly reworded your warning messages, but otherwise, this is on master now.

PR #112, opened 2016-10-28
doc: Don't delete flex.1 during "make distclean"

Explorer09:
flex.1 is pre-generated in release tarball. If we delete it, the next
"configure and make" on the source directory will then require help2man
unnecessarily.


westes:
Thanks! It's on master now.

Issue #99, opened 2016-10-25
Add configure option to not build libfl (Feature request)


Explorer09:
Copied from SourceForge [bug report #200](https://sourceforge.net/p/flex/bugs/200/) (Reported my me also)

Building flex without libfl is useful when you know that this flex build of yours will be used only with a cross-toolchain. For cross-building environments, having a libfl built for the host architecture is not useful. (We need one for the target architecture instead.)

For my case, although I would need to configure and build flex twice for a cross environment, I need to use a hack to disable libfl currently:

```
sed -i.bak -e 's/lib_LTLIBRARIES = \$/lib_LTLIBRARIES =/' \
    -e 's/\tlibfl.la \$/#&/' -e 's/\tlibfl_pic.la$/#&/' src/Makefile.in
```

It's better to make it into a configure option. (For your reference, GNU Bison can disable building liby by the '--disable-yacc' option.)

Alternatively, add a configure-time variable to explicitly specify which compiler will be used to build libfl (could be named `CC_FOR_TARGET` or `CC_FOR_LIBFL`). If using this approach, don't forget the respective `CFLAGS_FOR_TARGET` and other variables.

PR #115, opened 2016-10-31
--disable-libfl configure option 

Explorer09:
Two commits in this pull request:

First commit adds --disable-libfl configure option as mentioned in GH-99. Note that I didn't add `$CC_FOR_TARGET` or `--enable-sysroot` or whatever because it would be better to keep it simple and leave those options for building a cross-compiler out.

Second commit is a draft note of building flex for a cross-toolchain. I mention how to build libfl for the target system (when flex's `configure` does not support `--target`) and install it to a so-called "sysroot" directory.

I'm not expecting flex maintainer to merge this PR directly. It's better to review my changes first. And by the way **I didn't yet update the NEWS file for this change** (so please update that for me).

### DIFF
--- a/CrossCompile.txt
+++ b/CrossCompile.txt
@@ -1,0 +1,37 @@
+Notes for building flex for cross-toolchain
+-------------------------------------------
+
+Flex generates scanner code for C and C++, and has no architecture-
+specific parts. Therefore flex's build system does not support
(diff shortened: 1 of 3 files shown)


westes:
I put the commit to make building libfl optional on master. I'm holding off on documenting how to cross compile on the theory that I'll be able to get in a fix that does so soonish.

PR #120, opened 2016-11-06
Fix unportable test(1) operator.

0-wiz-0:
"==" is only supported by bash, "=" is the standard comparison
operator.


westes:
Thanks!

PR #122, opened 2016-11-08
Fixes a major bug in Flex's own lexing of literals

### LINKED ISSUE #113: v2.6.2 introduces behavior change that is incompatible with previous versions
With version 2.6.1 if prefix was set, then all affected functions would have their declaration changed and all references in the lexer input file would be changed as well.  With v2.6.2 only the function declarations are changed.

For example, set prefix to AA, reference yyget_extra() in the lexer input file, and build the C file using v2.6.1 and all calls in the C source file will be to AAget_extra().  However, build with v2.6.2 and the resultant C file will still have calls to yyget_extra() resulting in a broken source file.

This change in behavior appears to have been introduced with commit ee87e5df0056671be04eefaaeb85979d7bfca89c.

DemiMarie:
My changes caused Flex to mishandle string and character literals
in line comments.  This commit fixes them and adds a test case.

Fixes #113 .

This should probably be backported to 2.6.2, since it is a regression.


westes:
I won't back port it since the practice has been that new releases supercede old releases, but there will be a 2.6.3 sooner rather than later.

westes:
Thanks. it's in.

PR #125, opened 2016-11-16
Let stage1flex respect LFLAGS.

Explorer09:
...when building stage1scan.c from scan.l, otherwise this additional
bootstrapping stage will not make too much sense.

This is trivial change.


westes:
Good catch. It's on master.

PR #126, opened 2016-11-18
Ditch libcompat.la. Link $(LIBOBJS) from src/ dir.

Explorer09:
This helps bootstrapping flex as we no longer need AR or RANLIB for
such a little library.

All of its source files (malloc.c realloc.c reallocarray.c) are now
linked individually from the src/ directory.

By the way, the use of libtool on libcompat was not even recommended in
Automake manual [1], before this ditching.

[1] https://www.gnu.org/software/automake/manual/html_node/LIBOBJS.html
   "Because '$(LIBOBJS)' and '$(ALLOCA)' contain object file names that
    end with '.$(OBJEXT)', they are not suitable for Libtool libraries
    (where the expected object extension is .lo): LTLIBOBJS and LTALLOCA
    should be used instead."

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -28,7 +28,7 @@ AC_INIT([the fast lexical analyser generator],[2.6.3],[flex-help@lists.sourcefor
 AC_CONFIG_SRCDIR([src/scan.l])
 AC_CONFIG_AUX_DIR([build-aux])
 LT_INIT
-AM_INIT_AUTOMAKE([-Wno-portability foreign check-news std-options dist-lzip parallel-tests 1.14.1])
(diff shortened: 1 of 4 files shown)


Explorer09:
Note: during the making of this pull request, I discovered a bug on Autoconf's side.
Here, the AC_FUNC_MALLOC and AC_FUNC_REALLOC checks are not aware of cross-compiling and does not emit a warning or try to workaround it.
This mean that rpl_malloc and rpl_realloc will always be used when cross-compiling flex, even though the malloc and realloc functions on `--host` machine are GNU libc's behavior.

westes:
I have put this on master, albeit after making a more explicit comment in lib/Makefile.am and rewriting your commit messages and adding a NEWS entry. Thanks.

PR #129, opened 2016-11-26
Build system tweaks and '--disable-bootstrap' configure option

Explorer09:
Some build system & configure script tweaks and a configure option.
See description of each commit.


westes:

PR #131, opened 2016-12-01
Disallow '[' and ']' in prefix & overquote it

### LINKED ISSUE #128: malformed input file can hang flex
The attached file will hang flex, probably causing it to be stuck in an infinite loop. It's only one line, so it can also easily be tested on the command line:
```
echo '%option 8bit prefix="yywrap%"' | flex
```

This was found with the fuzzing tool american fuzzy lop. Tested with the current git code of flex.
[flex-hang.zip](https://github.com/westes/flex/files/603125/flex-hang.zip)

DemiMarie:
Stop the infinite loop in #128 by overquoting the `M4_YY_PREFIX` macro, so that it is always treated as a literal by M4.  Also reject `[` or `]` in it (which would have caused a C syntax error anyway).

Fixes #128.

### DIFF
--- a/src/buf.c
+++ b/src/buf.c
@@ -166,7 +166,7 @@ struct Buf *buf_strdefine (struct Buf *buf, const char *str, const char *def)
  */
 struct Buf *buf_m4_define (struct Buf *buf, const char* def, const char* val)
 {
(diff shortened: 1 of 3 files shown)


westes:
This is now on master and will be included in the next release.

Issue #49, opened 2016-01-08
flex-2.6.0 introduced bison dependency


0-wiz-0:
Since 2.6.0, flex now needs bison during build.
Before that, at least on NetBSD, yacc was enough.

The error I see is

```
gmake[1]: Entering directory '/scratch/devel/flex/work/flex-2.6.0/tests'
/bin/sh ../build-aux/ylwrap bison_nr_parser.y y.tab.c bison_nr_parser.c y.tab.h `echo bison_nr_parser.c | sed -e s/cc$/hh/ -e s/cpp$/hpp/ -e s/cxx$/hxx/ -e s/c++$/h++/ -e s/c$/h/` y.output bison_nr_parser.output -- yacc -d -p test
yacc: e - line 81 of "/scratch/devel/flex/work/flex-2.6.0/tests/bison_nr_parser.y", syntax error
        if( $1 != @1.first_line || $1 != testget_lineno())
                  ^
Makefile:2075: recipe for target 'bison_nr_parser.c' failed
gmake[1]: *** [bison_nr_parser.c] Error 1
../src/flex -o bison_nr_scanner.c bison_nr_scanner.l
/bin/sh ../build-aux/ylwrap bison_yylloc_parser.y y.tab.c bison_yylloc_parser.c y.tab.h `echo bison_yylloc_parser.c | sed -e s/cc$/hh/ -e s/cpp$/hpp/ -e s/cxx$/hxx/ -e s/c++$/h++/ -e s/c$/h/` y.output bison_yylloc_parser.output -- yacc -d -p test
yacc: e - line 83 of "/scratch/devel/flex/work/flex-2.6.0/tests/bison_yylloc_parser.y", syntax error
        if( $1 != @1.first_line || $1 != testget_lineno(scanner))
                  ^
Makefile:2075: recipe for target 'bison_yylloc_parser.c' failed
gmake[1]: *** [bison_yylloc_parser.c] Error 1
../src/flex -o bison_yylloc_scanner.c bison_yylloc_scanner.l
/bin/sh ../build-aux/ylwrap bison_yylval_parser.y y.tab.c bison_yylval_parser.c y.tab.h `echo bison_yylval_parser.c | sed -e s/cc$/hh/ -e s/cpp$/hpp/ -e s/cxx$/hxx/ -e s/c++$/h++/ -e s/c$/h/` y.output bison_yylval_parser.output -- yacc -d -p test
updating bison_yylval_parser.h
```

[...]

PR #144, opened 2016-12-31
test: Remove BUILT_SOURCES from tests/Makefile

Explorer09:
This will prevent anything in tests/ from being built when using the default "make all" target.
Partial fix for #49 (the "get everything out of BUILT_SOURCES in the test suite" part)

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -339,15 +339,24 @@ FLEX = $(top_builddir)/src/flex
 
 bison_nr_main.$(OBJEXT): bison_nr_parser.h bison_nr_scanner.h
 bison_nr_scanner.$(OBJEXT): bison_nr_parser.h
+
 bison_nr_scanner.h: bison_nr_scanner.c
+	@if test ! -f $@; then rm -f $<; else :; fi
+	@if test ! -f $@; then $(MAKE) $(AM_MAKEFLAGS) $<; else :; fi
 
 bison_yylloc_main.$(OBJEXT): bison_yylloc_parser.h bison_yylloc_scanner.h
 bison_yylloc_scanner.$(OBJEXT): bison_yylloc_parser.h
+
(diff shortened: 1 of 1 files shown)


westes:
A couple of questions:

I generally don't like silencing make commands. Especially in the flex test suite, since building these programs and testing with a non-$PATH flex is so non-standard anyway. So why did you use the @ to suppress the output?

And why test for the nonexistence of the files to begin with? That's make's job -- as well as checking for up-to-date-ness, too.

Explorer09:
@westes 
No, I didn't come up with the idea to silence `make` in lines like that. I just copy what automake did.
The lines are there because, for example, `header_r_scanner.h` doesn't directly derived from `header_r_scanner.c`, rather, it's generated from `header_r_scanner.l`. If `header_r_scanner.h` does not exist but `header_r_scanner.c` does, `make` can be confused and stop. (I admit this problem is rare.) Automake resolves this by silently deleting and remaking the file -- the same code I present in the commit. Perhaps automake people think these two lines are so trivial so that they suppress the output (the line that invokes flex or ylwrap will still be printed anyway). I don't know.

If you don't like the @ symbol then you may stop using it. Just change to something like this:

```
if test ! -f $@; then rm -f $<; fi
if test ! -f $@; then $(MAKE) $(AM_MAKEFLAGS) $<; fi
```

Or this:

```
if test ! -f $@; then rm -f $< && $(MAKE) $(AM_MAKEFLAGS) $<; fi
```

westes:
That's actually a good reason, so thanks for explaining why you did it. I'll have another look and see what I think. You're right that that edge case is rare, but given that this is the flex test suite, accounting for it sooner rather than later is a good thing.

westes:
Thanks. This is now on master and will be included in the next release of flex.

PR #146, opened 2017-01-01
Fix 'make indent' target

Explorer09:
This 'make indent' target has not been working for long since the
directories reorganization in flex 2.6.0. Now make it work again.
Unfortunately the current indent profile breaks many styles of existing
code.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -29,7 +29,6 @@
 # to DEFS.
 
 ACLOCAL_AMFLAGS = -I m4
-indent = @INDENT@
 
 dist_doc_DATA = \
 	AUTHORS \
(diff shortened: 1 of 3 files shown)


westes:
Agreed. the .indent.pro is inconsistant with itself even. I'm definitely open for a pr against the contents of the file itself. Although perhaps that should be hashed out on the flex-devel mailing list.

Makefile.am
```diff
@@ -58,8 +57,11 @@ ChangeLog: $(srcdir)/tools/git2cl
 		$(srcdir)/tools/git2cl > $@ \
 	; fi
 
+indent:
+	cd src && $(MAKE) $(AM_MAKEFLAGS) indent
```
westes: Shouldnt the cd src reference $(top_srcdir) because vpath builds?

Explorer09: No, this one isn't for that. This "cd src" is to run the "indent" target from the "src/Makefile", which then cd's to $(top_srcdir). You know this "indent" target is non-recursive, right? So if we run "make indent" in top_srcdir, let it descend to "src" directory to do the job.

Explorer09: And by the way, there won't exist $(top_srcdir)/src/Makefile in vpath build. It's actually $(top_builddir)/src/Makefile that is run.

westes:

PR #148, opened 2017-01-01
test: Correct tests/README about "skip" exit status.

Explorer09:


westes:
Thanks. This is now on master and will be included in the next release.

That was a holdover from the original implementation of the test suite, btw.

PR #149, opened 2017-01-02
testwrapper.sh fix against 'one' test error & removing test scripts bashisms

Explorer09:
Two changes are in this pull request:
The first change allows a so-called "one" test to exit with a non-zero status. `testwrapper.sh` will respect the status and no longer go further to compare the outputs.
The second change removes bashisms in the testwrapper scripts so that, for example, `dash` can be used to run `make check` now.


westes:

PR #150, opened 2017-01-02
(Experimental) dist-hook to force remake scan.c

Explorer09:
Ensure the flex scanner included in release tarball being generated by exactly the flex version to be released.

(Because I found out that the scan.c included in flex-2.6.3 release is generated by flex-2.6.2)

(I label this pull request experimental because I'm not sure how automake's "make dist" rule exactly work, and I don't have the exact build environment as the maintainer's to test that completely.)

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -92,6 +92,10 @@ stage1scan.c: scan.c
 	sed 's|^\(#line .*\)"'`printf %s $< | sed 's|[][\\\\.*]|\\\\&|g'`'"|\1"$@"|g' $< > $@
 endif
 
+dist-hook: scan.l flex$(EXEEXT)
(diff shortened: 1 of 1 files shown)


westes:
Something like this is needed, yes. It's in the queue.

westes:
Thanks. This is on master with some modification and will be included in the next release.

westes:
So unfortunately automake is choking on this.

westes:
I had forgot to run make distcheck which -- correctly -- notes that src/scan.c has been rebuilt. The fix is probably to ensure that src/scan.c in the directory is rebuilt, but that is going to take some work. I would agree that the distributed src/scan.c should be built with current flex, though.

westes:
But we'll have to open a new issue, it looks like since there's no option to reopen this pr.

Issue #73, opened 2016-05-20
the path to m4


westes:
Add an option, say, --with-m4 to configure that allows a user to specify which m4 flex should use. This option, when passed, should cause a symbol to be set in config.h. src/main.c should use that symbol to call m4. If that option is not passed, then configure should do its best to find a suitable m4 and set the config.h symbol as previously described.

In particular, this effects Solaris, where /usr/bin/m4 is not suitable for use with flex and where GNU m4 is not in a standard place.

Explorer09:
Um... what is this issue report about? Isn't `./configure M4=/path_to_your_gnu_m4/m4` already sufficient?

westes:
No, this is about the path to the runtime m4.

On Monday,  2 January 2017, 11:28 pm -0800, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> Um... what is this issue report about? Isn't `./configure M4=/path_to_your_gnu_m4/m4` already sufficient?
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/westes/flex/issues/73#issuecomment-270062236

-- 
Will Estes
westes575@gmail.com

Explorer09:
I still don't get it. The current config.h already `#define`'s the M4 symbol (1). That is what will be compiled into flex code. And flex honors the M4 environment variable if set (2). What else do we lack then?

(1): In configure.ac, AC_DEFINE_UNQUOTED([M4], ...)
(2): In src/main.c, the getenv("M4") part of code.

westes:
The compile time and runtime m4 paths are not necessarily the same. That's the issue. Yes, one work around is  to use the environment variable.

There might also be some stuff about using a better check for which m4 we're using. Not all m4 are useable from flex's point of view.

There's a third issue that we probably want to internalize macro processing anyway since we don't use a lot of m4 features. (But that's covered in another issue.)

On Tuesday,  3 January 2017,  5:42 am -0800, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> I still don't get it. The current config.h already `#define`'s the M4 symbol (1). That is what will be compiled into flex code. And flex honors the M4 environment variable if set (2). What else do we lack then?
> 
> (1): In configure.ac, AC_DEFINE_UNQUOTED([M4], ...)
> (2): In src/main.c, the getenv("M4") part of code.
> 
> -- 
> You are receiving this because you authored the thread.
[...]

PR #153, opened 2017-01-06
build: support cross compiling

### LINKED ISSUE #78: flex 2.6.1 does not cross compile

hobbitalastair:
Check for cross compiling. If cross compiling, build stage1flex using a custom link command.
We also override LDADD since that adds the replacement implementations that are cross compiled, and instead always use the replacement library implementations.

We don't use BUILD_OBJEXT and BUILD_EXEEXT since it seems that automake does not support these.

Fixes #78.

Signed-off-by: Alastair Hughes <hobbitalastair@gmail.com>


rossburton:
With this patch to get a non-cross build working I need to set stage1flex_LINK = $(LINK) in the non-cross case, otherwise it tries using stage1flex_LINK which isn't defined.  I guess I can moan at libtool for this, but apart from that this appears to work for me.

hobbitalastair:
@rossburton thanks! - I've pushed an updated patch that should fix that. It seems to build fine locally now.

Explorer09:
I would like to take some time to look at this and test it.
But, I hope you could also look at issue #73 and address it (have a build-time m4 path) in your cross-compile patch.

hobbitalastair:
@Explorer09 I'd rather keep that in a separate patch since cross-compiling does not require it.

Explorer09:
The BUILD_EXEEXT and BUILD_OBJEXT parts of AX_PROG_CC_FOR_BUILD is currently broken.
AX_PROG_CC_FOR_BUILD depend on AC_EXEEXT to determine the extensions, but the recent autoconf (2.69 in my testing) makes AC_EXEEXT expands to empty code so that BUILD_OBJEXT et al. will not be detected correctly. (Autoconf has internal _AC_COMPILER_EXEEXT macro, but it's supposed to be expanded once, hence it cannot substitute for AC_EXEEXT directly for AX_PROG_CC_FOR_BUILD use case.)
**This is a bug and should be reported.**

westes:
As per github, This branch has conflicts that must be resolved". Please resolve them and push the updates.Thanks!

westes:
Much thanks for the update. It's in the queue to review.

westes:
I had to remove the listing of lib/reallocarray.c since that file is no longer in the git tree.

westes:
In general, I don't like listing the lib/files explicitly because it means tracking them manually. That being said, I don't see a way around that.

westes:
I also note that ./.gitignore ignores m4/ and so I want to make sure that adding in the file in m4/ is not going to bite us. That is the right place for it, I am just concerned about the use of .gitignore above it.

westes:

PR #156, opened 2017-01-09
Update scan.l

luistung:
rm '\n' otherwise it cause 'bad character' error  when execute '/bin/sh ../build-aux/ylwrap scan.l lex.yy.c scan.c -- flex'


westes:
This is on master and will be included in the next release.

PR #161, opened 2017-01-12
filter: Don't emit #line directive to header if noline option is set

tklauser:
One place emitting a #line directive to the generated header was missed
in commit 647a92b9f4 when resolving #55. Fix it to respect gen_line_dirs
as well.


westes:
Thanks. THis is on master and will be included in the next release of flex.

PR #163, opened 2017-01-19
Add overflow detection for [re]allocate_array()

Explorer09:
- Use reallocarray() when we have it (i.e. in OpenBSD system). When we
  don't, use equivalent overflow detection for our allocate_array and
  reallocate_array functions.

- Remove lib/reallocarray.c from our LIBOBJS as we no longer need it.

- Provide a fallback SIZE_MAX macro definition in flexint.h (not
  preprocessor friendly, but enough for our reallocate_array use case).

P.S. I really wonder why we have to constrain the "num of elements"
argument to "int" size. The code doesn't seem to detect negative values,
and the limit of 32-bit "int" size seems to be unnecessary for a 64-bit
system. I would convert all of the uses to use size_t argument, if I have
time to try that.


westes:
> P.S. I really wonder why we have to constrain the "num of elements"
argument to "int" size. The code doesn't seem to detect negative values,
and the limit of 32-bit "int" size seems to be unnecessary for a 64-bit
system. I would convert all of the uses to use size_t argument, if I have
time to try that.

See my latest comment on #167. It's a holdover from POSIX. switching to size_t is not precisely wrong, but it does need accounting for given POSIX (and the --posix and related options).

westes:
I don't understand why we want this. reallocarray is clearly wanted -- your changes stil use it when present. So why not just supply it when absent as before?

Explorer09:
> westes 
> I don't understand why we want this. reallocarray is clearly wanted -- your changes stil [sic] use it when present. So why not just supply it when absent as before?

1. `reallocarray.c` from OpenBSD exposes some implementation details that could be unnecessary for flex use cases, hurts portability, and ... we can't make sure if it's better in performance.

For example a technical detail:

    #define MUL_NO_OVERFLOW ((size_t)1 << (sizeof(size_t) * 4))

and then...

    if ((nmemb >= MUL_NO_OVERFLOW || size >= MUL_NO_OVERFLOW) && nmemb > 0 && SIZE_MAX / nmemb < size) {

A naive overflow check will be:

    if (nmemb > 0 && SIZE_MAX / nmemb < size) {

I can doubt whether the former MUL_NO_OVERFLOW constant is really necessary at all. This should be a compiler detail: A smart compiler should be able to detect this construct and replace the 'div' instructions with a 'mul' and a carry flag jump.

Also, it sets errno which is not (yet) needed by flex use case. (Compiler will be hesitant to optimize out that one because errno is a global variable.)

And it's bad that it didn't provide a SIZE_MAX workaround when stdint.h (or inttypes.h) is not available for a system. (I'm experimenting with a minimal config.h for bootstrap which will assume stdint.h and inttypes.h not being available.)

2. Flex has allocate_array() and reallocate_array() functions, which are the only places that we will use reallocarray() from BSD libc. It'll be very unlikely that libc reallocarray() being used in anywhere else given that flex has its own interfaces.

westes:
That makes sense, thanks.

On Monday, 23 January 2017,  8:55 pm -0800, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> > westes 
> > I don't understand why we want this. reallocarray is clearly wanted -- your changes stil [sic] use it when present. So why not just supply it when absent as before?
> 
> 1. `reallocarray.c` from OpenBSD exposes some implementation details that could be unnecessary for flex use cases, hurts portability, and ... we can't make sure if it's better in performance.
> 
> For example a technical detail:
> 
>     #define MUL_NO_OVERFLOW ((size_t)1 << (sizeof(size_t) * 4))
> 
> and then...
> 
>     if ((nmemb >= MUL_NO_OVERFLOW || size >= MUL_NO_OVERFLOW) && nmemb > 0 && SIZE_MAX / nmemb < size) {
> 
> A naive overflow check will be:
> 
>     if (nmemb > 0 && SIZE_MAX / nmemb < size) {
> 
> I can doubt whether the former MUL_NO_OVERFLOW constant is really necessary at all. This should be a compiler detail: A smart compiler should be able to detect this construct and replace the 'div' instructions with a 'mul' and a carry flag jump.
> 
> Also, it sets errno which is not (yet) needed by flex use case. (Compiler will be hesitant to optimize out that one because errno is a global variable.)
> 
> And it's bad that it didn't provide a SIZE_MAX workaround when stdint.h (or inttypes.h) is not available for a system. (I'm experimenting with a minimal config.h for bootstrap which will assume stdint.h and inttypes.h not being available.)
> 
> 2. Flex has allocate_array() and reallocate_array() functions, which are the only places that we will use reallocarray() from BSD libc. It'll be very unlikely that libc reallocarray() being used in anywhere else given that flex has its own interfaces.
> 
> 
> -- 
> https://github.com/westes/flex/pull/163#issuecomment-274709778

-- 
Will Estes
westes575@gmail.com

westes:
This is on master and will be included in the next release.

PR #165, opened 2017-01-21
Fix void pointer arith in buf_append()

Explorer09:
Minor patch.

AFAIK, void pointer arithmetic is undefined behavior in standard C. GCC allowed it as an extension but `-Wpointer-arith` will warn it.


Explorer09:
Closing this as it's included in #167.

Issue #62, opened 2016-02-27
Architect flex types


westes:
A lot of compiler warnings are generated by the current mess of variable types in flex. In general, this means two things:

* We haven't taken a wholistic approach to "how big can things get?" which is more relevant now that we're well into a 64bit era.
* We're careless re signed and unsigned comparisons.

To consider this issue resolved, we need:

* A lexer that compiles under some set of options like:

-Werror -Wtype-limits -Wconversion

cleanly. There are likely other flags that are relevant as well.

* An approach that accounts for the fact that the various integer types can be of varying sizes. In particular, we'd like to let things grow as much as we can without imposing any arbitrary limits that we don't have to.

* POSIX imposes things like yyleng should be an int. It's not so much that that is wrong (but it is), but that people (rightly) expect yyleng to be an int, so we need a way to work with that expectation. In an ideal world, we'd provide "here's the crappy thing you expect because the standard says so and we get that" as well as "here's the bigger better thing that lets you work up to the limits of what a lexer written in C can do".

fuzxxl:
yyleng not being an int affects me, too. Perhaps it might be useful to have flex obey `POSIXLY_CORRECT` or to behave in compatibility mode by default when invoked as `lex` instead of `flex`.

westes:
There are some things turned on by POSIXLY_CORRECT, and by -l. But the type of yyleng is a bit harder to wrangle. It's not just yyleng, but all the variables it gets compared to and used in calculations with as well.

On Tuesday, 28 February 2017,  6:54 am -0800, Robert Clausecker <notifications@github.com> wrote:

> yyleng not being an int affects me, too. Perhaps it might be useful to have flex obey `POSIXLY_CORRECT` or to behave in compatibility mode by default when invoked as `lex` instead of `flex`.
> 
> -- 
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub:
> https://github.com/westes/flex/issues/62#issuecomment-283060500

-- 
Will Estes
westes575@gmail.com

PR #167, opened 2017-01-22
NetBSD fixes for flex-2.6.3

0-wiz-0:
I've adapted Christos' changes against flex-2.6.3 to git head.
Except for the two commits that are mine, please direct all questions to him.

This supersedes pull request #166.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -76,7 +76,7 @@ CLEANFILES = stage1scan.c stage1flex$(EXEEXT)
 MAINTAINERCLEANFILES = skel.c
 
 skel.c: flex.skl mkskel.sh flexint.h tables_shared.h tables_shared.c
-	sed 's/4_/a4_/g; s/m4preproc_/m4_/g' flex.skl | \
+	sed 's/4_/a4_/g; s/m4preproc_/m4_/g' $(srcdir)/flex.skl | \
 	  $(m4) -P -I $(srcdir)						   \
 	    -DFLEX_MAJOR_VERSION=`   echo $(VERSION)|cut -f 1 -d .`	   \
 	    -DFLEX_MINOR_VERSION=`   echo $(VERSION)|cut -f 2 -d .`	   \
@@ -86,7 +86,7 @@ skel.c: flex.skl mkskel.sh flexint.h tables_shared.h tables_shared.c
 
 if ENABLE_BOOTSTRAP
 stage1scan.c: scan.l stage1flex$(EXEEXT)
-	./stage1flex$(EXEEXT) $(AM_LFLAGS) $(LFLAGS) -o $@ $<
+	./stage1flex$(EXEEXT) $(AM_LFLAGS) $(LFLAGS) -o $@ $(srcdir)/scan.l
 else
 stage1scan.c: scan.c
 	sed 's|^\(#line .*\)"'`printf %s $< | sed 's|[][\\\\.*]|\\\\&|g'`'"|\1"$@"|g' $< > $@
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -76,7 +76,7 @@ CLEANFILES = stage1scan.c stage1flex$(EXEEXT)
 MAINTAINERCLEANFILES = skel.c
 
 skel.c: flex.skl mkskel.sh flexint.h tables_shared.h tables_shared.c
-	sed 's/4_/a4_/g; s/m4preproc_/m4_/g' $< | \
+	sed 's/4_/a4_/g; s/m4preproc_/m4_/g' flex.skl | \
 	  $(m4) -P -I $(srcdir)						   \
 	    -DFLEX_MAJOR_VERSION=`   echo $(VERSION)|cut -f 1 -d .`	   \
 	    -DFLEX_MINOR_VERSION=`   echo $(VERSION)|cut -f 2 -d .`	   \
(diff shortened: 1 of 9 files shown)


Explorer09:
I'll close pull request #165 ("Fix void pointer arith in buf_append()") because of this.

Also, may I comment something?
1. (commit f950f2b7eae99dcbb7e915193c4ff32b532df314) Why use strlcpy? It's non-portable and you didn't provide a workaround for systems that don't have strlcpy (e.g. glibc).
2. (commit deb594364b5c80f48771d63e7ba932698ce10b5b) Should use `$(srcdir)/flex.skl` otherwise it will break VPATH build.

westes:
> (commit f950f2b) Why use strlcpy? It's non-portable and you didn't provide a workaround for systems that don't have strlcpy (e.g. glibc).

BSD folks like it and for good reason. I think it's possible to provide a copy in lib/ and have the licenses play nicely.

> (commit deb5943) Should use $(srcdir)/flex.skl otherwise it will break VPATH build.

Yup, this sort of thing is a good catch.

Explorer09:
Well, just now I have been reading the articles regarding why glibc rejects strlcpy [1][2], and I feel their concern makes sense: strlcpy can cause silent data truncation if not checked carefully.

So what is the main motivation of this change? To prefer data truncation to buffer overflow or what?
By the way, the following can be considered equivalent if not considering the return value...

    strlcpy( nmstr, yytext, sizeof(nmstr) );
    memcpy( nmstr, yytext, strlen(yytext)+1 < sizeof(nmstr) ? strlen(yytext)+1 : sizeof(nmstr) );

[1] https://lwn.net/Articles/507319/
[2] https://stackoverflow.com/questions/2114896/

westes:
Yes, buffer overflows are a security flaw on top of being a bug. data truncation, not so much.

On Sunday, 22 January 2017,  6:34 pm -0800, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> Well, just now I have been reading the articles regarding why glibc rejects strlcpy [1][2], and I feel their concern makes sense: strlcpy can cause silent data truncation if not checked carefully.
> 
> So what is the main motivation of this change? To prefer data truncation to buffer overflow or what?
> By the way, the following can be considered equivalent if not considering the return value...
> 
>     strlcpy( nmstr, yytext, sizeof(nmstr) );
>     memcpy( nmstr, yytext, strlen(yytext)+1 < sizeof(nmstr) ? strlen(yytext)+1 : sizeof(nmstr) );
> 
> [1] https://lwn.net/Articles/507319/
> [2] https://stackoverflow.com/questions/2114896/
> 
> -- 
> https://github.com/westes/flex/pull/167#issuecomment-274384370

-- 
Will Estes
westes575@gmail.com

Explorer09:
I've read the code a bit further and I can declare that all strlcpy use in commit f950f2b7eae99dcbb7e915193c4ff32b532df314 are **totally unnecessary**.

Really. Both `nmdef` and `nmstr` arrays are declared with length `MAXLINE`, and before their real `strcpy` / `strlcpy` calls, all have done proper length checks.

So the use of strlcpy here benefits nothing but hurting compatibility.

Side note: strlcpy(dest,src,n) has a weakness by design that it's always time bound by O(src) because its return value always requires a strlen(src) internally. For cases that truncation is a desired result, a O(n) time will be better.

westes:
I'll have a look at each commit. In the past, I've turned down strlcpy stuff. It's a BSD-ism that is done not so much for specific benefits as trying to level set the playing field in a certain way. We don't particularly desire truncation, we just don't desire overflows.

0-wiz-0:
Would making them bounded using strncpy be fine?

westes:
I expect so. The issue is that without an strlcpy() implementation, we have to reject code that uses that function since, e.g., on linux, there is no strlcpy().

On Monday, 23 January 2017,  7:21 am -0800, Thomas Klausner <notifications@github.com> wrote:

> Would making them bounded using strncpy be fine?
> 
> -- 
> https://github.com/westes/flex/pull/167#issuecomment-274516419

-- 
Will Estes
westes575@gmail.com

westes:
Since the commits are on such different parts of flex, I'm cherry picking and putting them on master as I work through them.

westes:
You can see my progress by watching the master and gh-167 branches.

westes:
So I've worked through everything except the int/yy_size_t stuff.

That needs more care. POSIX defines, for example, yyleng as int, and folks expect that.

I'd probably be good with something that took POSIX into account and that defined int/yy_size_t accordingly. POSIX gets int, the contemporary world gets yy_size_t.

That gives folks a work around if they really need one.

Otherwise, it's all on master in nice clean commits.

So thanks to the both of you and the rest of the NetBSD team.

westes:
Closing this pull request. We can track the int/yy_size_t issue separately, as a part of #62 .

Explorer09:
Sorry for commenting too late, but I think strncpy could be a little bit worse in performance because strncpy has a _padding_ behavior. If we really want a better length-checked function, I would suggest strncat with a wrapper:

    static inline flex_strncpy(char *dest, const char* src, size_t n) {
        /* Like strncpy but always zero-terminate the string and without a padding. */
        if (n>0) {
            dest[0] = '\0';
            return strncat(dest, src, n - sizeof('\0'));
        } else {
            return dest;
        }
    }

But I'm sticking to my original position for now: strcpy is fine; no additional bound check needed.

@0-wiz-0 Are you aware of this?

PR #181, opened 2017-02-03
Fix flex_regcomp's awkward error reporting (double malloc).

Explorer09:
Fix unneeded double malloc - let regerror() write its message after our
"regcomp for (regex) failed: " string.

Also, errbuf cannot be free()'d because of flexfatal containing a
longjmp. (Too bad that flex didn't use gcc's
`__attribute__((noreturn))` or else we can catch this earlier.)


westes:

PR #183, opened 2017-02-07
flex.skl: fix indent (use always 4spaces)

alagoutte:


westes:
I'm inclined to wait on this sort of thing and to do it via a sed one liner when the time comes. In particular, this sort of cleanup will make using the -w switch with git blame mandatory for a lot of operations.

alagoutte:
Hi Westes,

Ok i understood... but it is before push some change... and avoid to make crazy my editor (and me) to known the good indent...

westes:
I'll do this sort of thing later -- There are a couple of releases of flex that need to happen in the meantime.

You can always configure your editor to use a 4-space tab stop as a work around. I do this in my own set up and it's easy enough to configure different source trees to behave differently as required.

PR #184, opened 2017-02-12
Faster way to determine power of two

Explorer09:
Replace the naive "for" loop in determining power of two with a clever
bitwise solution. This code is around the Internet already and is in
Public Domain.


westes:

PR #187, opened 2017-02-18
Fix code will never be executed [-Wunreachable-code]

alagoutte:
When build with clang (and use Wunreachable-code CFLAG)

### DIFF
--- a/src/flex.skl
+++ b/src/flex.skl
@@ -1363,11 +1363,6 @@ do_action:	/* This label is used only to access EOF actions. */
 					goto do_action;
 					}
 
-				else
-					{
-					if ( ! YY_G(yy_did_buffer_switch_on_eof) )
-						YY_NEW_FILE;
-					}
 				break;
 				}
 


Explorer09:
Wait. How is that part of code unreachable? And how are you sure that deleting that part of code is safe?
It look like non-trivial code.

alagoutte:
if you look the line 1349 ( yywrap( M4_YY_CALL_ONLY_ARG ) the code generated is xxxxwrap(yyscanner )

and from src/main.c (line 1590)
out_str ("\n#define %swrap(yyscanner) (/*CONSTCOND*/1)\n", prefix);
-> Always 1 (and else case is never execute...)

Explorer09:
On Mon, Feb 20, 2017 at 2:27 AM, Alexis La Goutte
<notifications@github.com> wrote:
>
> if you look the line 1349 ( yywrap( M4_YY_CALL_ONLY_ARG ) the code generated is xxxxwrap(yyscanner )
>
> and from src/main.c (line 1590)
> out_str ("\n#define %swrap(yyscanner) (/CONSTCOND/1)\n", prefix);
> -> Always 1 (and else case is never execute...)

It's wrapped (no pun intended) in (!do_yywrap) conditional.
That is, you can only assume it's always 1 if "%option yywrap" is unset.

Sorry, but your fix is incorrect.

westes:
Closing as per comments.

alagoutte:
Ok, i will search for make a better fix :)

Issue #133, opened 2016-12-09
flex should not always generate code that calls exit


taphqyu:
Hey there, thanks for maintaining flex! I noticed that the default YY_FATAL_ERROR macro calls exit(). This doesn't feel right for something that's often going to be built into a low-level part of another program -- there should be a way to report the error out, and have the containing program decide what to do.

I don't see a way to provide an alternate YY_FATAL_ERROR definition right now because the macro is called in many different places with different semantics. For example, it's called from different helper functions that all return different data types, relying on the fact that YY_FATAL_ERROR never returns.

Would it be possible to refactor the fatal error reporting so that a user of flex can avoid calling exit()? For example, if I could provide a YY_FATAL_ERROR definition that let me store the error message in my yyextra, and then return 0 from yylex(), that would meet my need. Thanks!

whydoubt:
As I had indicated, this is related to #194 .  However, I believe this issue requires extra work to fix.

Explorer09:
Dammit. No! #133 and #194 are different!
#133 is about calling exit in _generated scanner_ code, while #194 is about exiting of flex generator itself.

westes:
Oops, you're quite right. I got caught up in the paperwork and was moving too quickly yesterday.

saschmit:
On the other hand, the lack of an enforced abort()/exit() caused Grub's YY_FATAL_ERROR handler to allow a buffer overflow: https://eclypsium.com/2020/07/29/theres-a-hole-in-the-boot/

PR #192, opened 2017-04-03
scanner: Clean up resources when done

whydoubt:
Running all the tests through valgrind revealed numerous locations where
pointers were being overwritten or simply not being freed.

After cleaning these up, the testsuite still runs cleanly.

### DIFF
--- a/src/buf.c
+++ b/src/buf.c
@@ -217,9 +217,20 @@ void buf_destroy (struct Buf *buf)
 	if (buf) {
 		free(buf->elts);
 		buf->elts = NULL;
+		buf->nelts = 0;
 	}
 }
 
+void buf_destroy_full(struct Buf * buf, DestroyFunc destroy)
+{
+	void **p;
+	if (!buf)
+		return;
+	for (p = (void **)buf->elts;
+	     p < (void **)buf->elts + buf->nelts; p++)
+		destroy(*p);
+	buf_destroy(buf);
+}
 
 /* appends ptr[] to buf, grow if necessary.
  * n_elem is number of elements in ptr[], NOT bytes.
--- a/src/main.c
+++ b/src/main.c
@@ -159,6 +159,11 @@ void flex_atexit (void)
 
 	free_sym_tables ();
 
+	free (outfilename);
+	free (headerfilename);
+	free (prefix);
+	free (yyclass);
+
 	/* Free everything allocated in flexinit */
 	free (action_array);
 
@@ -406,7 +411,7 @@ void check_options (void)
 			snprintf (outfile_path, sizeof(outfile_path), outfile_template,
 				 prefix, suffix);
 
-			outfilename = outfile_path;
+			outfilename = xstrdup(outfile_path);
 		}
 
 		prev_stdout = freopen (outfilename, "w+", stdout);
@@ -1038,7 +1043,7 @@ void flexinit (int argc, char **argv)
 	reentrant = bison_bridge_lval = bison_bridge_lloc = false;
 	performance_report = 0;
 	did_outfilename = 0;
-	prefix = "yy";
+	prefix = xstrdup("yy");
 	yyclass = 0;
 	use_read = use_stdout = false;
 	tablesext = tablesverify = false;
@@ -1215,12 +1220,14 @@ void flexinit (int argc, char **argv)
 			break;
 
 		case OPT_OUTFILE:
-			outfilename = arg;
+			free (outfilename);
+			outfilename = xstrdup (arg);
 			did_outfilename = 1;
 			break;
 
 		case OPT_PREFIX:
-			prefix = arg;
+			free (prefix);
+			prefix = xstrdup (arg);
 			break;
 
 		case OPT_PERF_REPORT:
@@ -1336,7 +1343,8 @@ void flexinit (int argc, char **argv)
 			break;
 
 		case OPT_HEADER_FILE:
-			headerfilename = arg;
+			free (headerfilename);
+			headerfilename = xstrdup (arg);
 			break;
 
 		case OPT_META_ECS:
@@ -1390,7 +1398,8 @@ void flexinit (int argc, char **argv)
 			break;
 
 		case OPT_YYCLASS:
-			yyclass = arg;
+			free (yyclass);
+			yyclass = xstrdup (arg);
 			break;
 
 		case OPT_YYLINENO:
--- a/src/main.c
+++ b/src/main.c
@@ -169,6 +169,8 @@ void flex_atexit (void)
 
 	flex_finalize_regex ();
 
+	close_input_file ();
+
 	buf_destroy (&userdef_buf);
 	buf_destroy_full (&defs_buf, (DestroyFunc)free);
 	buf_destroy (&yydmap_buf);
--- a/src/main.c
+++ b/src/main.c
@@ -157,6 +157,8 @@ void flex_atexit (void)
 			free (dfaacc[i].dfaacc_set);
 	}
 
+	free_sym_tables ();
+
 	/* Free everything allocated in flexinit */
 	free (action_array);
 
--- a/src/main.c
+++ b/src/main.c
@@ -518,6 +518,8 @@ void check_options (void)
 
 		if (yytbl_hdr_fwrite (&tableswr, &hdr) <= 0)
 			flexerror (_("could not write tables header"));
+
+		yytbl_hdr_finalize (&hdr);
 	}
 
 	if (skelname && (skelfile = fopen (skelname, "r")) == NULL)
--- a/src/main.c
+++ b/src/main.c
@@ -133,8 +133,12 @@ const char *escaped_qend   = "]]M4_YY_NOOP]M4_YY_NOOP]M4_YY_NOOP[[";
 /* For debugging. The max number of filters to apply to skeleton. */
 static int preproc_level = 1000;
 
+static char *m4_path = NULL;
+
 void flex_atexit (void)
 {
+	free (m4_path);
+
 	/* Free everything allocated in flexinit */
 	free (action_array);
 
@@ -400,13 +404,11 @@ void check_options (void)
 	    char *slash;
 		m4 = M4;
 		if ((slash = strrchr(M4, '/')) != NULL) {
-			m4 = slash+1;
 			/* break up $PATH */
 			const char *path = getenv("PATH");
-			if (!path) {
-				m4 = M4;
-			} else {
-				int m4_length = strlen(m4);
+			if (path) {
+				int m4_length = strlen(slash+1);
+				m4 = slash+1;
 				do {
 					size_t length = strlen(path);
 					struct stat sbuf;
@@ -415,19 +417,18 @@ void check_options (void)
 					if (!endOfDir)
 						endOfDir = path+length;
 
-					{
-						char *m4_path = calloc(endOfDir-path + 1 + m4_length + 1, 1);
+					m4_path = calloc(endOfDir-path + 1 + m4_length + 1, 1);
 
-						memcpy(m4_path, path, endOfDir-path);
-						m4_path[endOfDir-path] = '/';
-						memcpy(m4_path + (endOfDir-path) + 1, m4, m4_length + 1);
-						if (stat(m4_path, &sbuf) == 0 &&
-							(S_ISREG(sbuf.st_mode)) && sbuf.st_mode & S_IXUSR) {
-							m4 = m4_path;
-							break;
-						}
-						free(m4_path);
+					memcpy(m4_path, path, endOfDir-path);
+					m4_path[endOfDir-path] = '/';
+					memcpy(m4_path + (endOfDir-path) + 1, m4, m4_length + 1);
+					if (stat(m4_path, &sbuf) == 0 &&
+						(S_ISREG(sbuf.st_mode)) && sbuf.st_mode & S_IXUSR) {
+						m4 = m4_path;
+						break;
 					}
+					free(m4_path);
+					m4_path = NULL;
 					path = endOfDir+1;
 				} while (path[0]);
 				if (!path[0])
@@ -445,6 +446,8 @@ void check_options (void)
     }
     filter_destroy_chain(output_chain);
     output_chain = NULL;
+    free(m4_path);
+    m4_path = NULL;
     yyout = stdout;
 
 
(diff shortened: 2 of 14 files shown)


Explorer09:
I don't like some of your changes here. They sound like bloat.
As an example, do we really need the `is_tablesfilename_alloc` flag variable?
If the `tablesfilename` pointer could never point to a static or stack buffer, then the value of NULL should be  enough to indicate it's unallocated. No need of an additional flag.

Also, would you please split up your changes to one kind of cleanup per commit? Otherwise I'll be very difficult to review your code, when you fuse several cleanup parts at once.

Explorer09:
By the way, boolean values in function arguments is a poor coding practice. If possible, use sparingly, or replace them with readable enums or bit fields.

whydoubt:
1) I have split up the patches, and will resubmit after retesting.
2) If I used an `_alloc` variable, there _is_ the possibility that the pointer can be from the stack.
3) I reworked the patch with the boolean argument to avoid doing so.

Explorer09:
@whydoubt 
> If I used an _alloc variable, there is the possibility that the pointer can be from the stack.

Then you need to prove it. AFAIK, attempting to call free() on a pointer to a stack could result in corruption of malloc's internal data (or [a segfault](https://stackoverflow.com/questions/4110767/free-function-without-malloc-or-calloc)). It should be **serious** error and potentially vulnerable to a local DoS (denial of service).

And yet you didn't demonstrate how to trigger such a thing. I would suggest it might be your misunderstanding of the code, rather than something serious. You know, not everything Valgrind reported are truly something you need to fix, and it's better you present us a *real* bug that could happen instead of "Valgrind says so".

src/buf.c
```diff
@@ -51,6 +51,7 @@ struct Buf defs_buf;		/**< for #define's autogenerated. List of strings. */
 struct Buf yydmap_buf;		/**< string buffer to hold yydmap elements */
 struct Buf m4defs_buf;          /**< m4 definitions. List of strings. */
 struct Buf top_buf;             /**< contains %top code. String buffer. */
+struct Buf alloc_buf;           /**< List of strings that need to be freed. */
```
Explorer09: @whydoubt  
Could you please explain why is this change needed? If I read correctly, buf_append() **does not** allocate additional buffer, but *reallocates* existing buffer instead.

You know how realloc() works, right? When realloc() succeeds, the data in the old buffer will be copied to new one, and the old buffer get free'd.

So I don't see the purpose of this `alloc_buf`  tracking all the things. Besides, I think your implementation is incorrect... (see below)

src/buf.c
```diff
@@ -266,4 +285,10 @@ struct Buf *buf_append (struct Buf *buf, const void *ptr, int n_elem)
 	return buf;
 }
 
+struct Buf *buf_append_alloc (struct Buf *buf, const void *ptr, int n_elem)
+{
+	buf_append(&alloc_buf, ptr, n_elem);
```
Explorer09: Here. What's the purpose of duplicating the data copied to alloc_buf?

src/main.c
```diff
@@ -109,6 +109,7 @@ int     nlch = '\n';
 
 bool    tablesext, tablesverify, gentables;
 char   *tablesfilename=0,*tablesname=0;
+bool    is_tablesfilename_alloc;
```
Explorer09: I've said before. Having a additional flag to track whether `tablesfilename` has been allocated doesn't make sense.
If `tablesfilename` will never point to a static or stack buffer, `tablesfilename == NULL` shall be sufficient to check that.

Also, [strdup()](http://pubs.opengroup.org/onlinepubs/9699919799/functions/strdup.html) **duplicates** the buffer the pointer point to, and the new buffer can be safely called free() against.
Please, have some common POSIX knowledge.

Your fix of this is incorrect.

src/parse.y
```diff
@@ -206,8 +206,14 @@ option		:  TOK_OUTFILE '=' NAME
 			{ yyclass = xstrdup(nmstr); }
 		|  TOK_HEADER_FILE '=' NAME
 			{ headerfilename = xstrdup(nmstr); }
-	    |  TOK_TABLES_FILE '=' NAME
-            { tablesext = true; tablesfilename = xstrdup(nmstr); }
```
Explorer09: To fix an xstrdup memory leak, the right way will be simply:
```
free(tablesfilename); /* Discard old buffer if user specifies --tables-file twice. */
tablesfilename = xstrdup(nmstr);
```

src/main.c
whydoubt: I said before, in the commit message, "Additionally, --tables-file=FILE will set tablesfilename to the argument **which should not be freed**, so this case is guarded against".

So there are three possible states for `tablesfilename`:
1. `NULL`
2. assigned from `--tables-file` command-line argument, which should be left alone because freeing it is undefined
3. strdup() of the `%option tables-file` value, which should be freed.
 
My fix is one reasonable solution.  Another would be to strdup() the command-line argument.

Explorer09: @whydoubt 
Could you please point to me in which part of the code can `tablesfilename` point to a non-allocated buffer.

src/parse.y
whydoubt: This is where the `%option tables-file` value is handled, not the command-line argument.
Perhaps this is where part of your misunderstanding stems from.  The command-line argument
is handled elsewhere, but my fix does not require any changes to that part of the code.

Explorer09: @whydoubt 
I apologize first for overlooking the "command-line argument" part of the code.
I found it in main.c:
```
		case OPT_TABLES_FILE:
			tablesext = true;
			tablesfilename = arg;
			break;
```
But still, having an additional `is_tablesfilename_alloc` doesn't convince me.
I'll suggest another way:
First, change the aforementioned code in main.c to `tablesfilename = xstrdup(arg);` that is, always strdup the name string before assigning to the global `tablesfilename` variable.
Then, for `%option tables-file` or additional `--tables-file` in the argument, check if `tablesfilename` has been assigned before. If yes, we can print a warning ("foo.l: warning: multiple specification of %option tables-file; ignoring all but last one"), and then you can do your memory leak fixes.

src/buf.c
whydoubt: Some `Buf`s wrap arrays of int's or char's.  However, others wrap arrays of `char*`s.  The _size_ of the array is what is realloc()ed. But the `char*` is just an entry in the array. Some are assigned from a const value and some from a strdup().  We cannot free all entries, because that would fault since some are const.
(Continued below)

westes:
I like this way of doing things.

On Friday,  7 April 2017,  7:11 am -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> 
> 
> 
> > @@ -206,8 +206,14 @@ option		:  TOK_OUTFILE '=' NAME
>  			{ yyclass = xstrdup(nmstr); }
>  		|  TOK_HEADER_FILE '=' NAME
>  			{ headerfilename = xstrdup(nmstr); }
> -	    |  TOK_TABLES_FILE '=' NAME
> -            { tablesext = true; tablesfilename = xstrdup(nmstr); }
> 
> @whydoubt 
> I apologize first for overlooking the "command-line argument" part of the code.
> I found it in main.c:
> ```
> 		case OPT_TABLES_FILE:
> 			tablesext = true;
> 			tablesfilename = arg;
> 			break;
> ```
> But still, having an additional `is_tablesfilename_alloc` doesn't convince me.
> I'll suggest another way:
> First, change the aforementioned code in main.c to `tablesfilename = xstrdup(arg);` that is, always strdup the name string before assigning to the global `tablesfilename` variable.
> Then, for `%option tables-file` or additional `--tables-file` in the argument, check if `tablesfilename` has been assigned before. If yes, we can print a warning ("foo.l: warning: multiple specification of %option tables-file; ignoring all but last one"), and then you can do your memory leak fixes.
> 
> -- 
> https://github.com/westes/flex/pull/192#discussion_r110398572

-- 
Will Estes
westes575@gmail.com

src/buf.c
whydoubt: (Continued from above)
So if the value is from a strdup(), calling buf_append_alloc adds the char* (NOTE: the pointer, not the string contents) both to the requested `Buf` _and_ to `alloc_buf`.  When we are done with the `Buf`s, `buf_destroy_alloc` will free `alloc_buf` _and each_ of its entries, i.e. the exact set of pointers that were strdup()ed for all the other `Buf`s.

src/parse.y
whydoubt: I agree, an extra strdup() would be better than all the extra is_xxx_alloc guards.

src/buf.c
whydoubt: Alternately, we could strdup() every string that we put in the buffer, but I'm not so sure that the resulting code will be much better that the solution I've presented.

Explorer09: So the problem you are trying to address is a `Buf` of allocated strings? (Similar to `std::vector<const char *>` if I use C++ terminology.)

I don't think your solution is the best way, because, what if I change a single `char` pointer (element) to point to another string without freeing the previous string? It's also memory leak that way.

I think the better way is to force all  `Buf` of strings instances to be responsible to their own string allocation. Don't use a global "things to free()" buffer. Instead implement proper finalize() or destructor for the `Buf` of strings case.

whydoubt: It is more that the contents of a `Buf` of `char*` can be a mixed bag of allocated vs const strings.

While `Buf` is somewhat like a std::vector, `Buf`'s interface only allows for appending entries, not for removing or changing them.

I was careful to limit alloc_buf use to buf.c, so the scope is actually file (though I did forget to set it static).
This actually seems like a very normal pattern if you think of buf_init_alloc as initializing the buf interface
buf_destroy_alloc as finalizing.

What you suggest will involve either altering the `Buf` structure to contain its own list of free-able elements,
or require every element be free-able (come from a strdup, etc.).  In the former case, there would still need to be buf_append_alloc.  In the latter case, we would either need change `Buf` to know about the element type (currently it just knows about the element size) maybe with an enum, or there would need to be a separate variant of buf_destroy() and if you want the interface to take care of the strdup then we still need buf_append_alloc.

Explorer09: whydoubt
> While Buf is somewhat like a std::vector, Buf's interface only allows for appending entries, not for removing or changing them.

> I was careful to limit alloc_buf use to buf.c, so the scope is actually file (though I did forget to set it static).
> This actually seems like a very normal pattern if you think of buf_init_alloc as initializing the buf interface
> buf_destroy_alloc as finalizing.

A class-wide alloc_buf is a bad idea when what you are trying to fix only matters to a particular _instantiation_ of a _generic_ type. In this case, I mean the `Buf` of allocated strings.

> What you suggest will involve either altering the Buf structure to contain its own list of free-able elements,
or require every element be free-able (come from a strdup, etc.). In the former case, there would still need to be buf_append_alloc. In the latter case, we would either need change Buf to know about the element type (currently it just knows about the element size) maybe with an enum, or there would need to be a separate variant of buf_destroy() and if you want the interface to take care of the strdup then we still need buf_append_alloc.

Actually, I was expecting the latter. If a Buf instance contains an array of strings, then either all of its strings shall be free-able or none at all. If a Buf instance are strings that should be free'd when destroyed, then do this subroutine:

```
void buf_strings_destroy (struct Buf *buf)
{
	char *p;
	if (!buf) return;
	for (p = buf->elts + (buf->nelts - 1) * buf->elt_size; p >= buf->elts; p -= buf->elt_size) {
		free(*p);
	}
	free(buf->elts);
	buf->elts = NULL;
}
```

You may consider this as subclassing---I just create a variant of the finalizer that calls free() on all of its members as well as freeing the array itself.

Explorer09:
Excuse me. I have one concern about the flex_atexit() function that @whydoubt introduces.
Normally when the program exits, all allocated memory will be automatically freed up by the operating system, so a routine to free them at exit isn't needed. Note that I'm not saying I'm against flex_atexit() function -- tracking and listing what should be freed in the end of main() is a good idea. But such routine is only useful when flex is being used as a library. So, instead of registering an atexit() for this, how about changing this routine to the opposite of set_up_initial_allocations()? That is, instead of allocating required structures, it destroys all of them and finalizes flex's global variables.

westes:
Agreed. It is conceivable that we will want to provide flex as a library at some point in the future.

On Saturday,  8 April 2017,  5:02 am -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> Excuse me. I have one concern about the flex_atexit() function that @whydoubt introduces.
> Normally when the program exits, all allocated memory will be automatically freed up by the operating system, so a routine to free them at exit isn't needed. Note that I'm not saying I'm against flex_atexit() function -- tracking and listing what should be freed in the end of main() is a good idea. But such routine is only useful when flex is being used as a library. So, instead of registering an atexit() for this, how about changing this routine to the opposite of set_up_initial_allocations()? That is, instead of allocating required structures, it destroys all of them and finalizes flex's global variables.
> 
> -- 
> https://github.com/westes/flex/pull/192#issuecomment-292713705

-- 
Will Estes
westes575@gmail.com

whydoubt:
Currently, flex calls exit() in 5 locations. Given this _current_ state of the code, atexit() seemed appropriate. Calling exit() is exceptionally bad practice for a library though, and would need to be cleaned up as part of making flex usable as a library. With no exit() calls, there should be no need to register with atexit(), and it could be called in a more natural location.

I could add a patch to the front of the patchset to change ~~all~~ most [see also #133] exit() calls to FLEX_EXIT(), then change the current first patch to name the function flex_finalize and call it from the `if (exit_status)` block in flex_main().

Explorer09:
@whydoubt 
I agree with you. Calling exit() directly in multiple places is a bad idea when flex is going to become usable as a library.
You can go ahead and fix all exit() calls to FLEX_EXIT(). But I suggest you make another pull request for it -- this one is already too many changes and too many commits. Let's have some changes that are easy to merge first.

westes:
Yes, to be clear, pull requests should be small and atomic.

On Sunday,  9 April 2017,  7:32 am -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> @whydoubt 
> I agree with you. Calling exit() directly in multiple places is a bad idea when flex is going to become usable as a library.
> You can go ahead and fix all exit() calls to FLEX_EXIT(). But I suggest you make another pull request for it -- this one is already too many changes and too many commits. Let's have some changes that are easy to merge first.
> 
> -- 
> https://github.com/westes/flex/pull/192#issuecomment-292789397

-- 
Will Estes
westes575@gmail.com

westes:
I'm going to close this. Please resubmit with things more atomic and taking into account additional changes to src/filter.c.

Thanks!

PR #194, opened 2017-04-10
FLEX_EXIT() is the preferred way to exit flex

whydoubt:


Explorer09:
I probably need to inform that all exit() calls you've changed are supposed to be called by child processes.
While this is not an error, FLEX_EXIT() could make an unneeded wait() call on the children, so I'm not sure whether your change would be useful in long term.

While you are at it, could you also look at the last `return 0;` line in `filter_tee_header` function?
A smart compiler might detect and warn that it's unreachable code.

whydoubt:
For the child processes, the while (wait(&child_status) > 0) block will not be entered because wait will return -1 as a result of having no children themselves.

main.c gives guidance in comments that "FLEX_EXIT(n) should be used instead of exit(n)". The usefulness in calling FLEX_EXIT() would be in consistency with this guidance, and so that there is a single place to add any future shutdown activities.

So should the code change to match the guidance, or should the guidance change to match the code?
Note that if we change the guidance, any future shutdown activities may need to be called before each exit() call.

whydoubt:
> While you are at it, could you also look at the last `return 0;` line in `filter_tee_header` function?
> A smart compiler might detect and warn that it's unreachable code.

Actually this exit() could be removed.  The result of the `return 0;` would be returning to where it was called:
```c
if ((r = chain->filter_func (chain)) == -1)
    flexfatal (_("filter_func failed"));
exit (0);
```
So the effect would be the same, but would remove one exit() from the code.
Also, the exit(1) is unreachable and can be removed as well.

Explorer09:
whydoubt
> For the child processes, the while (wait(&child_status) > 0) block will not be entered because wait will return -1 as a result of having no children themselves.
> main.c gives guidance in comments that "FLEX_EXIT(n) should be used instead of exit(n)". The usefulness in calling FLEX_EXIT() would be in consistency with this guidance, and so that there is a single place to add any future shutdown activities.
> So should the code change to match the guidance, or should the guidance change to match the code?
Note that if we change the guidance, any future shutdown activities may need to be called before each exit() call.

That's a problem regarding policy that I cannot give you advice. Maybe @westes can answer this one.
Anyway, I think the fact that flex may fork itself to run filters could bring problem if flex has to be usable as a library. There are some pull requests that tried to replace M4 dependency for something else, but that would be a major change that can't be done short-term, I guess.

> Actually this exit() could be removed. The result of the return 0; would be returning to where it was called:

What I like is consistency. All child routines should have the same way to exit and return back to the caller thread. If flex had been using libpthread (instead of forking), pthread_exit() would be the ideal call for this.
But I can guess FLEX_EXIT() was never meant to called by child processes or threads when it was designed. Heck, this inconsistency could trouble me, too.

westes:
> @whydoubt
> > For the child processes, the while (wait(&child_status) > 0) block will not be entered because wait will return -1 as a result of having no children themselves.
> > main.c gives guidance in comments that "FLEX_EXIT(n) should be used instead of exit(n)". The usefulness in calling FLEX_EXIT() would be in consistency with this guidance, and so that there is a single place to add any future shutdown activities.
> > So should the code change to match the guidance, or should the guidance change to match the code?
> Note that if we change the guidance, any future shutdown activities may need to be called before each exit() call.

Please change the code to match the guidance.

> Anyway, I think the fact that flex may fork itself to run filters could bring problem if flex has to be usable as a library. There are some pull requests that tried to replace M4 dependency for something else, but that would be a major change that can't be done short-term, I guess.

Those pull requests were doing a ton of other things as well as removing m4. Ultimately, we should do all our work internally and remove the m4 dependency as it introduces complexity and we barely scratch the surface of what m4 does anyway.

> What I like is consistency. All child routines should have the same way to exit and return back to the caller thread. If flex had been using libpthread (instead of forking), pthread_exit() would be the ideal call for this.
> But I can guess FLEX_EXIT() was never meant to called by child processes or threads when it was designed. Heck, this inconsistency could trouble me, too.

We have to be mindful of portability as well. We can't just assume that libpthread is available to us, iirc.

whydoubt:
> Please change the code to match the guidance.

OK. That is what this pull request does.

westes:

westes:
Fixes #133

PR #195, opened 2017-04-10
myesc() function fixes

Explorer09:


westes:

PR #196, opened 2017-04-12
flex.skl: add missing update of yy_buf_size after yyrealloc()

millert:
An alternate approach would be to allocate an extra 2 bytes in the yyrealloc() and set yy_buf_size to new_size.  The code in flex.skl is not consistent in this regard; sometimes an extra 2 bytes are allocated, other times yy_buf_size is adjusted by -2.


Explorer09:
Please describe the exact problem that you are fixing. Is this a real bug that can be demonstrated?
You know, a patch is not a substitute for a proper bug report.

westes:
Thanks!

On Friday, 14 April 2017,  7:54 am -0700, "Todd C. Miller" <notifications@github.com> wrote:

> There is a case in yy_get_next_buffer() where YY_CURRENT_BUFFER_LVALUE->yy_ch_buf is expanded but the buffer size is not updated to match the newly allocated size.  Because yy_buf_size is not updated, it could expand the buffer again every time  yy_get_next_buffer() is called.
> 
> I noticed this from inspection.
> 
> -- 
> https://github.com/westes/flex/pull/196#issuecomment-294167836

-- 
Will Estes
westes575@gmail.com

westes:

PR #198, opened 2017-05-02
minor improvements for invocation of build tools

GitMensch:
autogen.sh:
* check for libtoolize / glibtoolize before anything else
* use LIBTOOLIZE environment var, if set

configure.ac:
* version number in `AM_INIT_AUTOMAKE` must come first (otherwise you get an error message for unrecognized option `dist-lzip`)

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -44,6 +44,8 @@ fi
 #worried. (Don't worry, we *do* have a ChangeLog, we just need the
 #Makefile first.)
 
-touch ChangeLog
+if ! test -f ChangeLog; then
+   touch ChangeLog
+f
 "$LIBTOOLIZE" --install --force
 autoreconf --install --force
--- a/autogen.sh
+++ b/autogen.sh
@@ -46,6 +46,6 @@ fi
 
 if ! test -f ChangeLog; then
    touch ChangeLog
-f
+fi
 "$LIBTOOLIZE" --install --force
 autoreconf --install --force
--- a/autogen.sh
+++ b/autogen.sh
@@ -45,5 +45,5 @@ fi
 #Makefile first.)
 
 touch ChangeLog
-LIBTOOLIZE --install --force
+"$LIBTOOLIZE" --install --force
 autoreconf --install --force
--- a/autogen.sh
+++ b/autogen.sh
@@ -24,10 +24,26 @@
 # If you see no configure script, then run ./autogen.sh to create it
 # and procede with the "normal" build procedures.
 
+# use LIBTOOLIZE, if set
+LIBTOOLIZE_ORIG="$LIBTOOLIZE";
+if test "x$LIBTOOLIZE" = "x"; then LIBTOOLIZE=libtoolize; fi
+
+# test libtoolize
+$LIBTOOLIZE --version 2>/dev/null
+if test "$?" -ne 0; then
+   LIBTOOLIZE=glibtoolize
+   $LIBTOOLIZE --version 2>/dev/null
+   if test "$?" -ne 0; then
+      echo "error: libtoolize not working, re-run with LIBTOOLIZE=/path/to/libtoolize"
+      echo "       LIBTOOLIZE is currently \"$LIBTOOLIZE_ORIG\""
+      exit 1
+   fi
+fi
+
 #if we pretend to have a ChangeLog, then automake is less
 #worried. (Don't worry, we *do* have a ChangeLog, we just need the
 #Makefile first.)
 
 touch ChangeLog
-libtoolize --install --force || glibtoolize --install --force
+LIBTOOLIZE --install --force
 autoreconf --install --force
--- a/configure.ac
+++ b/configure.ac
@@ -28,7 +28,7 @@ AC_INIT([the fast lexical analyser generator],[2.6.4],[flex-help@lists.sourcefor
 AC_CONFIG_SRCDIR([src/scan.l])
 AC_CONFIG_AUX_DIR([build-aux])
 LT_INIT
-AM_INIT_AUTOMAKE([-Wno-portability foreign check-news std-options dist-lzip parallel-tests subdir-objects 1.14.1])
+AM_INIT_AUTOMAKE([1.14.1 -Wno-portability foreign check-news std-options dist-lzip parallel-tests subdir-objects])
 AC_CONFIG_HEADER([src/config.h])
 AC_CONFIG_LIBOBJ_DIR([lib])
 AC_CONFIG_MACRO_DIR([m4])
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -1,32 +1,14 @@
-help2man = @HELP2MAN@
-
 FLEX = $(top_builddir)/src/flex$(EXEEXT)
 
 info_TEXINFOS =	flex.texi
 dist_man_MANS = flex.1
 MAINTAINERCLEANFILES = flex.1
 
-CLEANFILES = \
-	flex.aux \
-	flex.cp \
-	flex.cps \
-	flex.fn \
-	flex.fns \
-	flex.hk \
-	flex.hks \
-	flex.ky \
-	flex.log \
-	flex.op \
-	flex.ops \
-	flex.pg \
-	flex.toc \
-	flex.tp \
-	flex.tps \
-	flex.vr \
-	flex.vrs
+CLEANFILES = *.aux *.cp *.cps *.fn *.fns *.hk *.hks *.ky *.log \
+	*.op *.ops *.pg *.toc *.tp *.tps *.vr *.vrs
 
 flex.1: $(top_srcdir)/configure.ac $(top_srcdir)/src/flex.skl $(top_srcdir)/src/options.c $(top_srcdir)/src/options.h | $(FLEX)
-	$(help2man) --name='$(PACKAGE_NAME)' --section=1 \
+	$(HELP2MAN) --name='$(PACKAGE_NAME)' --section=1 \
 	--source='The Flex Project' --manual='Programming' \
 	--output=$@ $(FLEX) \
 	|| rm -f $@


GitMensch:
note: I don't see a "lzip" release tarball yet. If these aren't used the option may should be removed from `configure.ac`. BTW: If this option is removed it works perfectly fine to generate the file with automake 1.11.

I suggest to lower the version to the first automake version that generates a nice configure + Makefiles which would be the first version that supports `dist-lzip` (I think it was 1.12, but I didn't found a release announcement including this information).
As long as `dist-lzip` is in I suggest to add a comment before that the version may be lowered to 1.11 if the dist-lzip option is removed.

westes:
The lzip is intentional and should be retained. It replaced the xz option due to concerns with that format.

On Tuesday,  2 May 2017,  3:40 am -0700, Simon Sobisch <notifications@github.com> wrote:

> note: I don't see a "lzip" release tarball yet. If these aren't used the option may should be removed from `configure.ac`. BTW: If this option is removed it works perfectly fine to generate the file with automake 1.11.
> 
> I suggest to lower the version to the first automake version that generates a nice configure + Makefiles which would be the first version that supports `dist-lzip` (I think it was 1.12, but I didn't found a release announcement including this information).
> As long as `dist-lzip` is in I suggest to add a comment before that the version may be lowered to 1.11 if the dist-lzip option is removed.
> 
> -- 
> https://github.com/westes/flex/pull/198#issuecomment-298600835

-- 
Will Estes
westes575@gmail.com

Explorer09:
If I remember correctly, `autoreconf` itself can run libtoolize when needed, ideally it's better to run libtoolize  through autoreconf, and we will have no need to check the existence of that tool by ourselves.

@GitMensch Could you try and see if that's possible?

GitMensch:
> The lzip is intentional and should be retained. It replaced the xz option due to concerns with that format.

OK, then the automake version should still be lowered to the minimal needed version for dist-lzip ;-) I've rechecked and the info page of automake 1.11.3 is the first that has the option `dist-lzip` listed. I'll add this to the repo with the pull request.

> If I remember correctly, autoreconf itself can run libtoolize when needed, ideally it's better to run libtoolize through autoreconf, and we will have no need to check the existence of that tool by ourselves.
>
>@GitMensch Could you try and see if that's possible?

Yes, I will and report back.

westes:
Yes we do need to run libtoolize manually. There's a bug where some things don't get picked up unless libtoolize is run first.

On Tuesday,  2 May 2017,  5:11 am -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> If I remember correctly, `autoreconf` itself can run libtoolize when needed, ideally it's better to run libtoolize  through autoreconf, and we will have no need to check the existence of that tool by ourselves.
> 
> @GitMensch Could you try and see if that's possible?
> 
> -- 
> https://github.com/westes/flex/pull/198#issuecomment-298617625

-- 
Will Estes
westes575@gmail.com

GitMensch:
... ok, I'm not looking at running libtoolize via autoreconf...

Explorer09:
By the way, the preference of lzip vs xz is (in my opinion) a political issue, and I would not bother with which one the maintainer will choose, as long as a fallback (e.g. gz) is available.

Off-topic, but I would say that either format has its pros and cons. lzip's merit is its simplicity, but it lacks feature that would be useful for binary packages (i.e. the BCJ filter). xz works very well with binary packages but the format is rather complex, and some criticize as being bloated. I have no preference of either, for now (although I have some concern about lzip's developer's attitude).

westes:
Yes, the intent is to keep gzip and lzip at the current time.

On Tuesday,  2 May 2017,  5:21 am -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> By the way, the preference of lzip vs xz is (in my opinion) a political issue, and I would not bother with which one the maintainer will choose, as long as a fallback (e.g. gz) is available.
> 
> Off-topic, but I would say that either format has its pros and cons. lzip's merit is its simplicity, but it lacks feature that would be useful for binary packages (i.e. the BCJ filter). xz works very well with binary packages but the format is rather complex, and some criticize as being bloated. I have no preference of either, for now (although I have some concern about lzip's developer's attitude).
> 
> -- 
> https://github.com/westes/flex/pull/198#issuecomment-298619734

-- 
Will Estes
westes575@gmail.com

Explorer09:
If we need to run libtoolize first before autoreconf, can we just use a simpler solution in autogen.sh?
From this:

    libtoolize --install --force || glibtoolize --install --force

To this:

   "${LIBTOOLIZE:-libtoolize}" --install --force || glibtoolize --install --force || exit 1

GitMensch:
I want to do my lasts commit for this pull request and want to ask if you see any issues in changing the version dependencies in configure.ac:

```diff
-AM_INIT_AUTOMAKE([1.14.1 -Wno-portability foreign check-news std-options dist-lzip parallel-tests subdir-objects])
-AM_GNU_GETTEXT_VERSION([0.19])
+AM_INIT_AUTOMAKE([1.11.3 -Wno-portability foreign check-news std-options dist-lzip parallel-tests subdir-objects])
+AM_GNU_GETTEXT_VERSION([0.18])
```
Using these versions `make distcheck` works as does `make dist-lzip`. I personally don't you any reason to enforce a newer version, do you?

GitMensch:
> If we need to run libtoolize first before autoreconf, can we just use a simpler solution in autogen.sh?

@Explorer09 
Is there something you don't like about https://github.com/GitMensch/flex/blob/master/autogen.sh?

westes:
Does git blame offer any insight when run against those lines?

I remember deliberately adding those versions, but I don't remember if it was just "I'm upgrading my tools" or if there was something specific.

That make distcheck passes does bode well as it's rather picky about what it does.

On Tuesday,  2 May 2017,  7:55 am -0700, Simon Sobisch <notifications@github.com> wrote:

> I want to do my lasts commit for this pull request and want to ask if you see any issues in changing the version dependencies in configure.ac:
> 
> ```diff
> -AM_INIT_AUTOMAKE([1.14.1 -Wno-portability foreign check-news std-options dist-lzip parallel-tests subdir-objects])
> -AM_GNU_GETTEXT_VERSION([0.19])
> +AM_INIT_AUTOMAKE([1.11.3 -Wno-portability foreign check-news std-options dist-lzip parallel-tests subdir-objects])
> +AM_GNU_GETTEXT_VERSION([0.18])
> ```
> Using these versions `make distcheck` works as does `make dist-lzip`. I personally don't you any reason to enforce a newer version, do you?
> 
> 
> -- 
> https://github.com/westes/flex/pull/198#issuecomment-298659976

-- 
Will Estes
westes575@gmail.com

westes:
So here's what happened...

We used to just run autoreconf.

Then we added libtool things. autoreconf is supposed to call libtool, but it didn't.

So we added a call to libtoolize directly because that fixes things.

Then, it turns out that OS X doesn't call it libtoolize. Apple specialness dictates it is called glibtoolize. So that was hacked in.

The right solution is probably to set libtoolize's variable to a default value, then test for libtoolize (and set the variable if it is on path). Then to test for glibtoolize (and set the variable if that is found) if libtoolize was not found.

Then just use the variable when it comes execution time.

On Tuesday,  2 May 2017,  8:08 am -0700, Simon Sobisch <notifications@github.com> wrote:

> > If we need to run libtoolize first before autoreconf, can we just use a simpler solution in autogen.sh?
> 
> @Explorer09 
> Is there something you don't like about https://github.com/GitMensch/flex/blob/master/autogen.sh?
> 
> -- 
> https://github.com/westes/flex/pull/198#issuecomment-298663966

-- 
Will Estes
westes575@gmail.com

Explorer09:
@GitMensch
> Is there something you don't like about https://github.com/GitMensch/flex/blob/master/autogen.sh?

Over-complication. When my one-line tweak can do the same as what you did, why should we use a more complicated solution? After all, what we need is to run libtoolize early, right?

westes:
It is probably ok to conditionally touch the ChangeLog file, yes.

On Tuesday,  2 May 2017,  8:30 am -0700, Simon Sobisch <notifications@github.com> wrote:

> > The right solution is probably to set libtoolize's variable to a default value, then test for libtoolize (and set the variable if it is on path). Then to test for glibtoolize (and set the variable if that is found) if libtoolize was not found. Then just use the variable when it comes execution time.
> 
> That's exactly what the file in the pull request does + exits with an information the user understands + allows to explicit tell the script which libtoolize to chose.
> 
> > Over-complication.
> 
> The suggested one-line tweak always touches the ChangeLog (I'd even vote for adding an additional file check and only touch if the file isn't existing) and prompt the user with "glibtoolize not found".
> 
> -- 
> https://github.com/westes/flex/pull/198#issuecomment-298670547

-- 
Will Estes
westes575@gmail.com

westes:
This is on master and will be included in the next release of flex.

GitMensch:
This is 2.6.4 milestone then, isn't it?

westes:
Yes.

And the only thing that is a blocker on that is that I'd like to automate signing the release tarballs. Because distribution folks would like to be able to verify that they have the sources they think they have.

On Tuesday,  2 May 2017, 12:10 pm -0700, Simon Sobisch <notifications@github.com> wrote:

> This is 2.6.4 milestone then, isn't it?
> 
> -- 
> You are receiving this because you modified the open/close state.
> https://github.com/westes/flex/pull/198#issuecomment-298731673

-- 
Will Estes
westes575@gmail.com

GitMensch:
Signing them is a good thing. Please report back the changes you actually did for the automation, I'd like to revisit them for GnuCOBOL... Thank you.

westes:
You can have a look in gnulib. There's stuff that you could take as inspiration. I'm having to be a bit more cautious because of licensing, though. The gnulib stuff does assume that you're in the gnu environment, but short of that it looks reasonable as a starting point.

On Tuesday,  2 May 2017, 12:20 pm -0700, Simon Sobisch <notifications@github.com> wrote:

> Signing them is a good thing. Please report back the changes you actually did for the automation, I'd like to revisit them for GnuCOBOL... Thank you.
> 
> -- 
> You are receiving this because you modified the open/close state.
> https://github.com/westes/flex/pull/198#issuecomment-298734403

-- 
Will Estes
westes575@gmail.com

PR #202, opened 2017-05-02
Translation tweaks

GitMensch:
* POTFILES.in: use generated C files instead flex/bison sources to prevent warnings from xgettext
* replaced Rules-getpo by extra script to manually update and rebuild all po files
* Makevars: updated to recent version, don't auto-rebuild po-files during `make dist`


westes:
This is on master and will be included in the next release of flex.

PR #207, opened 2017-05-02
Don't leak macro definitions into header files

### LINKED ISSUE #142: "Macro redefined" warnings with multiple_scanners_r (in testsuite)
```bash
$ ./autogen.sh && ./configure && make
$ cd tests && make multiple_scanners_r
```
I can get multiple "macro redefined" warnings from this:

```
gcc -DHAVE_CONFIG_H -I. -I../src  -I../src -I../src   -g -O2 -MT multiple_scanners_r_main.o -MD -MP -MF .deps/multiple_scanners_r_main.Tpo -c -o multiple_scanners_r_main.o multiple_scanners_r_main.c
In file included from multiple_scanners_r_main.c:25:0:
multiple_scanners_r_2.h:20:0: warning: "yy_create_buffer" redefined [enabled by default]
 
 ^
In file included from multiple_scanners_r_main.c:24:0:
multiple_scanners_r_1.h:20:0: note: this is the location of the previous definition
 
 ^
In file included from multiple_scanners_r_main.c:25:0:
multiple_scanners_r_2.h:22:0: warning: "yy_delete_buffer" redefined [enabled by default]
 
 ^
In file included from multiple_scanners_r_main.c:24:0:
multiple_scanners_r_1.h:22:0: note: this is the location of the previous definition
 
 ^
In file included from multiple_scanners_r_main.c:25:0:
multiple_scanners_r_2.h:24:0: warning: "yy_scan_buffer" redefined [enabled by default]
 
 ^
[...]
```

This seems to be a regression introduced in 347652c32b4614995acd4ee0d686499da2070d9e "fix backwards incompatible changes in 2.6.2."

DemiMarie:
This allowed unnamespaced definitions to leak into header files,
breaking client code.

Fixes #142

### DIFF
--- a/src/flex.skl
+++ b/src/flex.skl
@@ -111,9 +111,10 @@ m4_ifdef( [[M4_YY_REENTRANT]],  [[m4_define([[M4_YY_HAS_START_STACK_VARS]])]])
 
 m4_ifdef( [[M4_YY_PREFIX]],, [[m4_define([[M4_YY_PREFIX]], [[yy]])]])
 
-m4preproc_define(`M4_GEN_PREFIX',
-    ``[[#define yy$1 ]]M4_YY_PREFIX[[$1]]
-%# m4_define([[yy$1]], [[M4_YY_PREFIX[[$1]]m4_ifelse($'`#,0,,[[($'`@)]])]])'')
+m4preproc_define(`M4_GEN_PREFIX',``
+[[#define yy$1 ]]M4_YY_PREFIX[[$1]]
+'m4preproc_divert(1)`
+[[#undef yy$1]]'m4preproc_divert(0)')
 
 %if-c++-only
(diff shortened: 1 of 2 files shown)


DemiMarie:
Yikes.  This *still* isn’t a full solution.

The real problem is that we need to have macro definitions that can process user code(!).  I think I can handle this though.

DemiMarie:
This is a partial solution.  It can still break if the user expects the old behavior of the `yy_*` macros in user header files not affecting the `yy_*` identifiers in the user's Lex file.

westes:
You should be able to test this against master without using --disable-nls now.

DemiMarie:
@westes Done.  The tests still pass.

westes:
This is on master and will be included in the next release. Much thanks.

PR #208, opened 2017-05-03
No need to generate lib/Makefile.in

Explorer09:
Files in lib/ are picked up and built using using makefile directives
in src/Makefile.am. Remove the need to generate lib/Makefile.in and the
dummy lib/Makefile.am.

(I discoved this tweak on my attempt to change flex to use "non-
recursive makefile". I'm not sure if flex should be migrated to use
non-recursive makefile, but it's good to get rid of lib/Makefile.in
first. Should flex use non-recursive makefile eventually? That's
my question. BTW, Bison's build system is non-recursive make.)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -42,7 +42,6 @@ EXTRA_DIST = \
 	autogen.sh
 
 SUBDIRS = \
-	lib \
 	src \
 	doc \
 	examples \
--- a/configure.ac
+++ b/configure.ac
@@ -175,7 +175,6 @@ doc/Makefile
 examples/Makefile
 examples/fastwc/Makefile
 examples/manual/Makefile
-lib/Makefile
 po/Makefile.in
 src/Makefile
 tools/Makefile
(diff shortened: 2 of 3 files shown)


westes:
Getting rid of lib/Makefile.in is fine since we pick up those files as needed in other ways.

What is the benefit of non-recursive make in this case?

Explorer09:
The benefit of non-recursive make:
* In general, non-recursive make can track dependencies across subdirectories better than recursive make.
* The size total of makefiles can be reduced.
The cost:
* Needs time to migrate and test.

westes:
Ah, then that is relevant to our interests.

It doesn't strike me as urgent, at least not relative to the next couple releases. So take your time with it.

On Wednesday,  3 May 2017,  7:25 am -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> The benefit of non-recursive make:
> * In general, non-recursive make can track dependencies across subdirectories better than recursive make.
> * The size total of makefiles can be reduced.
> The cost:
> * Needs time to migrate and test.
> 
> -- 
> https://github.com/westes/flex/pull/208#issuecomment-298926652

-- 
Will Estes
westes575@gmail.com

westes:

PR #209, opened 2017-05-03
Handle zero and negative bracecounts appropriately

DemiMarie:
Negative bracecounts, and actions terminating while in state `ACTION_STRING` or `CHARACTER_CONSTANT`, can happen on valid input, because Flex doesn't parse `//` comments or C++11 raw string literals.


westes:

PR #210, opened 2017-05-03
Delete comments from skeleton file

DemiMarie:
It's nearly trivial to do, and it saves space in the produced
executable.


westes:

PR #211, opened 2017-05-04
flex.skl: Mention %# comments are removed

Explorer09:
This is a follow-up of pull request #210.
It's good to mention the comment removal in flex.skl.


westes:

PR #213, opened 2017-05-04
Correct C code sample within documentation.

viktor-shepel:
**Issue:**
Documentation contains C code sample which generate compile time error.

**Root cause:**
Typographical error.

**Solution:**
Code sample provided with missing equality operator.

Ticket - https://github.com/westes/flex/issues/214


westes:

PR #216, opened 2017-05-06
-C*f and -C*F option-specific test files collide on case-insensitive

HBBroeker:
-C*f and -C*F option-specific test files collide on case-insensitive
file systems, and cause tests to either not be run at all, or to
overwrite each other's files at build time.  So rename -C*F ones to -C*_F.


westes:

Issue #219, opened 2017-05-17
flex-2.6.4: segfault during build


0-wiz-0:
When building flex-2.6.4 on NetBSD-7.99.71/amd64, it dumps core:
````
gmake[1]: Entering directory '/scratch/devel/flex/work/flex-2.6.4/src'
/usr/pkg/bin/gmake  all-am
gmake[2]: Entering directory '/scratch/devel/flex/work/flex-2.6.4/src'
./stage1flex   -o stage1scan.c ./scan.l
Makefile:1696: recipe for target 'stage1scan.c' failed
gmake[2]: *** [stage1scan.c] Segmentation fault (core dumped)
````
I'm not sure how far the backtrace is trustworty, since it looks like this:
````
Core was generated by `stage1flex'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  flexinit (argc=argc@entry=4, argv=argv@entry=0x7f7fff7d03f0) at main.c:976
976             action_array[0] = '\0';
(gdb) bt
#0  flexinit (argc=argc@entry=4, argv=argv@entry=0x7f7fff7d03f0) at main.c:976
#1  0x0000000000416a89 in flex_main (argc=4, argv=0x7f7fff7d03f0) at main.c:168
#2  0x0000000000401acb in ___start ()
#3  0x00007f7e35c03382 in _rtld (sp=<optimized out>, relocbase=<optimized out>) at /disk/6/archive/foreign/src/libexec/ld.elf_so/rtld.c:516
#4  0x00007f7fff7d0e41 in ?? ()
#5  0x00007f7fff7d0e4e in ?? ()
#6  0x00007f7fff7d0e6e in ?? ()
#7  0x00007f7fff7d0e7c in ?? ()
#8  0x00007f7fff7d0e9a in ?? ()
#9  0x00007f7fff7d0ea5 in ?? ()
#10 0x00007f7fff7d0edc in ?? ()
#11 0x00007f7fff7d0eea in ?? ()
#12 0x00007f7fff7d0f23 in ?? ()
#13 0x00007f7fff7d0f67 in ?? ()
#14 0x00007f7fff7d0f73 in ?? ()
#15 0x00007f7fff7d0f8d in ?? ()
#16 0x00007f7fff7d0f9a in ?? ()
#17 0x00007f7fff7d0fa7 in ?? ()
#18 0x00007f7fff7d0fed in ?? ()
#19 0x00007f7fff7d0ff8 in ?? ()
[...]

PR #221, opened 2017-05-19
Use reallocarr() when available.

0-wiz-0:
Fixes https://github.com/westes/flex/issues/219


westes:

PR #223, opened 2017-05-21
scanner: add optionn to specify backup filename


westes:
Currently, it is possible to request a backup file of lexer
states. Add an option to specify the filename for that backup so that
it is possible to do this with multiple lexers in the same
directory. This will also allow writing the backup file to somewhere
else in case the directory where the lexer is run is not writeable.


westes:
@EricSharkey The assignment to backing_name in src/main.c on line 1040 is invalide because backing_name is a const. Can you have a look and add a fix for that to this pull request?

EricSharkey:
On Sun, May 21, 2017 at 4:31 PM, Will Estes <notifications@github.com>
wrote:

> The assignment to backing_name in src/main.c on line 1040 is invalide
> because backing_name is a const. Can you have a look and add a fix for that
> to this pull request?
>

I see.

This code has changed since I wrote the patch against 2.5.35.  It was
declared as:

  static char *backing_name = "lex.backup";

and is currently declared as:

  static const char backing_name[] = "lex.backup";

so the const keyword is new, and the type has changed from char * to char
[].  The problem isn't really the const qualifier, it's the change from
pointer to arrray.

How would you like this fixed?  To me, it looks like the right fix is to go
back to a pointer instead of an array, but keep the new const qualifier.

  static const char *backing_name = "lex.backup";

Is that acceptable to you?

Eric

westes:
Yes.

On Monday, 22 May 2017,  6:58 am -0700, EricSharkey <notifications@github.com> wrote:

> On Sun, May 21, 2017 at 4:31 PM, Will Estes <notifications@github.com>
> wrote:
> 
> > The assignment to backing_name in src/main.c on line 1040 is invalide
> > because backing_name is a const. Can you have a look and add a fix for that
> > to this pull request?
> >
> 
> I see.
> 
> This code has changed since I wrote the patch against 2.5.35.  It was
> declared as:
> 
>   static char *backing_name = "lex.backup";
> 
> and is currently declared as:
> 
>   static const char backing_name[] = "lex.backup";
> 
> so the const keyword is new, and the type has changed from char * to char
> [].  The problem isn't really the const qualifier, it's the change from
> pointer to arrray.
> 
> How would you like this fixed?  To me, it looks like the right fix is to go
> back to a pointer instead of an array, but keep the new const qualifier.
> 
>   static const char *backing_name = "lex.backup";
> 
> Is that acceptable to you?
> 
> Eric
> 
> 
> -- 
> You are receiving this because you were assigned.
> https://github.com/westes/flex/pull/223#issuecomment-303107852

-- 
Will Estes
westes575@gmail.com

westes:
This is on master with additional changes submitted by the author of the patch. It will apear in the next release.

PR #224, opened 2017-05-23
Clean up unreachable code/unused variables.

0-wiz-0:
Requested by @Explorer09.

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -149,7 +149,8 @@ void   *allocate_array (int size, size_t element_size)
 			   ("memory allocation failed in allocate_array()"));
 
 	return mem;
-#elif HAVE_REALLOCARRAY
+#else
+#if HAVE_REALLOCARRAY
 	/* reallocarray has built-in overflow detection */
 	mem = reallocarray(NULL, (size_t) size, element_size);
 #else
@@ -162,6 +163,7 @@ void   *allocate_array (int size, size_t element_size)
 			   ("memory allocation failed in allocate_array()"));
 
 	return mem;
+#endif
 }
 
 
@@ -665,13 +667,14 @@ char   *readable_form (int c)
 
 void   *reallocate_array (void *array, int size, size_t element_size)
 {
-	void *new_array;
 #if HAVE_REALLOCARR
 	if (reallocarr(&array, (size_t) size, element_size))
 		flexfatal (_("attempt to increase array size failed"));
 
 	return array;
-#elif HAVE_REALLOCARRAY
+#else
+	void *new_array;
+#if HAVE_REALLOCARRAY
 	/* reallocarray has built-in overflow detection */
 	new_array = reallocarray(array, (size_t) size, element_size);
 #else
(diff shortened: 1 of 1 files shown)


Explorer09:
Wait. I think there's one more minor issue:

```
void   *reallocate_array (void *array, int size, size_t element_size)
{
#if HAVE_REALLOCARR
	if (reallocarr(&array, (size_t) size, element_size))
		flexfatal (_("attempt to increase array size failed"));

	return array;
#else
// ......
#endif
}
```

I don't think it's wise to simply pass `&array` to reallocarr(). I would be better if you do the copy first; then pass the copied variable address:

```
	void *new_array = array;
	if (reallocarr(&new_array, (size_t) size, element_size))
		flexfatal (_("attempt to increase array size failed"));

	return array;
```

The reason is that `array` may not be always passed in stack. If I use `regparm(3)` attribute from the compiler, the argument may be passed through a register instead. This can remove confusion for compiler, at least.

0-wiz-0:
I don't understand why this could be a problem. Do you have a pointer for more information on this?

Explorer09:
@0-wiz-0 
> I don't understand why this could be a problem. Do you have a pointer for more information on this?

Well, no. It's not really a "problem" by itself, but a suggestion so the code can be more readable and maintainable.

I've learned some calling conventions of CPU architectures, and I know it's a bad idea to simply pass an address of the function argument, because you won't know how the compiler will interpret that case when it comes to calling conventions. For x86-32, it's fine because arguments are passed on stack; for x86-64 ABI, however, first 6 arguments are passed in registers, and the compiler will have to workaround (e.g. by forcing that argument to be on stack) if the argument's address is taken. In short, this hinders compiler's optimization as compiler has less freedom on allocating registers. That's why I say taking addresses of function arguments is bad idea.

westes:
@Explorer09 but your solution just passes the address of a different array. So why is that an improvement since the same behavior exists -- passing the address of an array as a function argument.

Explorer09:
@westes, I think you have misunderstood about the interface of reallocarr().

reallocarr() is called differently than OpenBSD's reallocarray(), and unfortunately it isn't obvious by reading reallocarr's prototype. Allow me to rewrite the prototypes so that they are more readable:

    int reallocarr(    _Inout_  void **ptr_to_ptr_to_array,
                       _In_     size_t num_elements,
                       _In_     size_t element_size);
    /* The type of first argument would be better stated as (void**),
       but it uses (void*) for some obscure technical reason.
       Internally it's casted to (void **) before de-referencing. */
    void *reallocarray(_In_opt_ void *ptr_to_array,
                       _In_     size_t num_elements,
                       _In_     size_t element_size);
    void *realloc(     _In_opt_ void *ptr_to_array,
                       _In_     size_t size);

So for `reallocarr`, it isn't to pass an address of an array, but to pass "an address of a pointer to the array". (If reallocation succeeds, `reallocarr` will modify the pointer variable directly. Unlike `realloc` or `reallocarray`, who will return the address of newly allocated array, without modifying the pointer variable.)

Before you correct your understanding about `reallocarr`, I can't understand your comment.

westes:
@Explorer09 So could you either add a commit onto this (if you can; I'm unsure how github permissions work in this case) or submit an additional pull request that has the changes you suggest?

westes:
We will be using the other submitted version.

Issue #222, opened 2017-05-19
shell portability issue in tableopts.sh


0-wiz-0:
When building current git head on NetBSD, I see:
````
/bin/sh ./tableopts.sh > ./tableopts.am
./tableopts.sh: 20: Syntax error: Bad substitution
````
Line 20 is:
````
            testname=tableopts_${kind}_${threading}${opt/%F/_F}.${kind}
````
I don't think the `${opt/%F/_F}` syntax is portable.

Another issue in the same corner: the generated file `tests/tableopts.am` is checked into the repository.

westes:
Here's the thinking behind tableopts.am being checked in:

It's needed to generate tests/Makefile.in. It is extremely unenlightening -- so much so that it can be scripted. It's long and in some sense may be fragile and so is worth separating out into its own file.

Although perhaps that is an argument for adding it to autogen.sh at the top level?

What does NetBSD use as /bin/sh? We do aim for pretty broad portability in the shell scripts, so I'm open to a pull request on that front.

On Friday, 19 May 2017,  1:26 am -0700, Thomas Klausner <notifications@github.com> wrote:

> When building current git head on NetBSD, I see:
> ````
> /bin/sh ./tableopts.sh > ./tableopts.am
> ./tableopts.sh: 20: Syntax error: Bad substitution
> ````
> Line 20 is:
> ````
>             testname=tableopts_${kind}_${threading}${opt/%F/_F}.${kind}
> ````
> I don't think the `${opt/%F/_F}` syntax is portable.
> 
> Another issue in the same corner: the generated file `tests/tableopts.am` is checked into the repository.
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/westes/flex/issues/222

-- 
Will Estes
westes575@gmail.com

0-wiz-0:
NetBSD has its own sh, see http://netbsd.gw.com/cgi-bin/man-cgi?sh+1+NetBSD-current
Basically the same as OpenBSD's and FreeBSD's sh(1).

westes:
Ah, the % operator is supported but not the / operator.

PR #225, opened 2017-05-24
test: use portable pattern substitution in tableopts.sh

Explorer09:
Fixes: #222


westes:
@0-wiz-0, can you confirm that this works on NetBSD?

westes:
The test suite builds and runs under linux (aka my laptop).

0-wiz-0:
Yes, the patch fixes the problem. Thanks, @Explorer09 !

westes:
This is now on master. Much thanks.

PR #228, opened 2017-06-04
test: echo 'set -v' before setting so in shell.

Explorer09:
This is intended to be a follow-up of commit 2c0152a0c33061c63c928279e9bfcf496e1178a1


westes:

PR #230, opened 2017-06-20
filter: memory leak free scanner postprocessing.

viktor-shepel:
**Issue:**
Scanner postprocessing leaks memory during correction of `#line`
directives values and generation of C header file.

**Root cause:**
`filter_fix_linedirs` and `filter_tee_header` functions are not
dispose allocated memory.

**Solution:**
Automatically reclaim affected memory by allocating it on stack
insted of heap. Stack allocation should not be a problem as its
only 512 bytes and there is no recursive calls.

ticket - https://github.com/westes/flex/issues/232

### DIFF
--- a/src/filter.c
+++ b/src/filter.c
@@ -230,8 +230,7 @@ int filter_tee_header (struct filter *chain)
 	 * header file at the same time.
 	 */
 
-	const int readsz = 512;
-	char   *buf;
+	char    buf[512];
 	int     to_cfd = -1;
 	FILE   *to_c = NULL, *to_h = NULL;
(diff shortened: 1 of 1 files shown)


src/filter.c
```diff
@@ -230,8 +230,7 @@ int filter_tee_header (struct filter *chain)
 	 * header file at the same time.
 	 */
 
-	const int readsz = 512;
-	char   *buf;
+    char   buf[512];
```
viktor-shepel: will correct in a moment (whitespaces VS tabs)

PR #237, opened 2017-07-08
Faster is_blank_line() implementation

Explorer09:
Introduce static is_blank_line() function in filter.c.
Remove regex_blank_line and use is_blank_line() in place of its role.

### DIFF
--- a/src/filter.c
+++ b/src/filter.c
@@ -326,7 +326,15 @@ int filter_tee_header (struct filter *chain)
 
 static bool is_blank_line (const char *str)
 {
-	return (regexec (&regex_blank_line, str, 0, NULL, 0) == 0);
+#if 1 /* faster */
+	while (isspace(*str))
+		str++;
+	return (*str == '\0');
+#endif
+#if 0 /* smaller */
+	char c;
+	return (sscanf(str, " %c", &c) < 1);
+#endif
 }
 
 /** Adjust the line numbers in the #line directives of the generated scanner.
--- a/src/filter.c
+++ b/src/filter.c
@@ -324,6 +324,11 @@ int filter_tee_header (struct filter *chain)
 	return 0;
 }
 
+static bool is_blank_line (const char *str)
+{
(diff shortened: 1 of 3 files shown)


src/filter.c
```diff
@@ -326,7 +326,15 @@ int filter_tee_header (struct filter *chain)
 
 static bool is_blank_line (const char *str)
 {
-	return (regexec (&regex_blank_line, str, 0, NULL, 0) == 0);
+#if 1 /* faster */
+	while (isspace(*str))
+		str++;
+	return (*str == '\0');
+#endif
+#if 0 /* smaller */
```
westes: What is the purpose of the #if 1 / #if 0?

Explorer09: These are just two implementations that you can choose one. Both are faster and take smaller code than the current regex approach.
If you think there is no need to keep two pieces of (functionally equivalent) code, feel free to delete one of them, and have the other into master.

westes:
I will pick one of them and declutter the #if 1/0.

On Friday, 14 July 2017,  7:30 pm -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> 
> 
> 
> > @@ -326,7 +326,15 @@ int filter_tee_header (struct filter *chain)
>  
>  static bool is_blank_line (const char *str)
>  {
> -	return (regexec (&regex_blank_line, str, 0, NULL, 0) == 0);
> +#if 1 /* faster */
> +	while (isspace(*str))
> +		str++;
> +	return (*str == '\0');
> +#endif
> +#if 0 /* smaller */
> 
> These are just two implementations that you can choose one. Both are faster and take smaller code than the current regex approach.
> If you think there is no need to keep two pieces of (functionally equivalent) code, feel free to delete one of them, and have the other into master.
> 
> -- 
> https://github.com/westes/flex/pull/237#discussion_r127574457

-- 
Will Estes
westes575@gmail.com

westes:
Thanks. This is on master -- with some cleanup.

PR #238, opened 2017-07-09
{re,}allocate_array code cleanup

Explorer09:
My version of the code cleanup similar to #224.

After this change, the allocate_array() and reallocate_array() will look almost similar. I don't like this. If possible in the future, I would like to turn allocate_array() into a wrapper that calls reallocate_array() so ultimately we will have something like this:
```c
static inline void *allocate_array (size_t size, size_t element_size) {
	return reallocate_array (NULL, size, element_size)
}
```
However the error message prevents me from making the change now. By the way, the size argument would better to use size_t type if the API could be changed.

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -141,27 +141,24 @@ void add_action (const char *new_text)
 
 void   *allocate_array (int size, size_t element_size)
 {
-	void *mem;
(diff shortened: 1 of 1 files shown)


westes:

PR #240, opened 2017-07-09
Ignore comment lines (%#) in skeleton files

jannick0:
In skeleton files comments are indicated by leading `%#` and when directly read in using `flex -S <skeleton.skl>` they should be ignored. Example: `flex.skl`. 

Amending commit 2f21edac99b5efc432417233e6e53326d630e08f which removed this conditional branch.


westes:

PR #242, opened 2017-07-17
#define BASENAME and get rid of dependency to libgen.h

jannick0:
I am struggling a battle with github ... hoping that this comes through as it should do.


westes:
Thanks, this is on master and will be included in the next release.

PR #243, opened 2017-07-18
mkskel.sh: fix EOL issue for CRLF systems

jannick0:
`sed` called by `mkskel.sh` complained when the input file (here: `scan.l`) is converted to CRLF. The fix makes `sed` digest both LF and CRLF.


westes:

PR #245, opened 2017-07-18
scanner: remove BASENAME(); don't strip path from program_name

Explorer09:
*Note: please test if help2man on flex still generates correct man page (i.e. no "../src/flex" or such) when building flex after this. I didn't yet test it.*

There's no technical need of stripping path from program_name. I think
the users should be fine if they see the path they use to invoke flex
is diagnostic messages and help texts.

Yes, users will see "Usage: ../flex [OPTIONS]..." now if they invoke
flex with the path "../flex".

The --version output has been changed so that the name field will be
always "flex" or "flex++". If the flex program has been renamed to
"lex" (for compatibility or other reason) this will allow identifying
the implementation name ("flex"). (And it's a recommended practice in
GNU Coding Standards)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -996,7 +996,7 @@ void flexinit (int argc, char **argv)
     flex_init_regex();
 
 	/* Enable C++ if program name ends with '+'. */
-	program_name = BASENAME(argv[0]);
+	program_name = argv[0];
 
 	if (program_name != NULL &&
 	    program_name[strlen (program_name) - 1] == '+')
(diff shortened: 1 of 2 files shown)


westes:
Why is the call to gettext in the help text unnecessary? Seems like non-english-reading users would want everything translated.

Explorer09:
@westes

>Why is the call to gettext in the help text unnecessary? Seems like non-english-reading users would want everything translated.

That is not what I said and not what I've done.
What I said is that the translations of the help text will need to be updated after this. And I have never removed a gettext call from the help text.

PR #248, opened 2017-08-06
build: always include config.h before stdlib.h

### LINKED ISSUE #247: rpl_* functions and config.h/stdlib.h interations causes segfault with modern GCC

hobbitalastair:
`scan.c` included `stdlib.h` before `config.h`. This caused problems when using the fallback `rpl_malloc` and `rpl_realloc` functions, since the `#define malloc rpl_malloc` happened after the function prototypes, so the functions were used with no prototype. This somehow worked with GCC 6.2.1 but emitted warnings; with GCC 7.1.1 this caused `stage1flex` to segfault.

To fix this, create `stage1scan.c` and `scan.c` using a custom rule which prepends the generated files to ensure that they include config.h.

I've tested this by generated a distribution tarball and compiling (both normally and with the rpl_* functions), so hopefully things work right off the bat. There may be a cleaner way to distribute `scan.template`; I'm no automake expert so I just picked a method that worked.

Fixes #247.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -99,18 +99,31 @@ skel.c: flex.skl mkskel.sh flexint.h tables_shared.h tables_shared.c
 	$(SHELL) $(srcdir)/mkskel.sh $(srcdir) $(m4) $(VERSION) > $@.tmp
 	mv $@.tmp $@
 
+scan.c: scan.l scan.template
+	$(LEX) $(AM_LFLAGS) $(LFLAGS) -o $@ $(srcdir)/scan.l
+	mv $@ $@.new
+	cat $(srcdir)/scan.template $@.new > $@
+	rm -f $@.new
+
 if ENABLE_BOOTSTRAP
-stage1scan.c: scan.l stage1flex$(EXEEXT)
+stage1scan.c: scan.l stage1flex$(EXEEXT) scan.template
 	./stage1flex$(EXEEXT) $(AM_LFLAGS) $(LFLAGS) -o $@ $(srcdir)/scan.l
+	mv $@ $@.new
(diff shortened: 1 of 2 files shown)


westes:
scan.template is not making it into the distribution tarball, which makes "make distcheck" fail when trying to build scan.c.

Explorer09:
I'm thinking if there's a way better workaround. Since the only problem is the lack of prototypes, how about this:
For building flex's own scanner, just forward declare the `rpl_malloc` and `rpl_realloc` prototypes despite whether we would be using them or not.
This will fix the error. The long term solution would be to remove reliance on glibc's malloc behavior, but on standard one, however that would require more testing then.

westes:
Could you submit that as a pull request?

On Sunday,  3 September 2017,  5:28 pm +0000, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> I'm thinking if there's a way better workaround. Since the only problem is the lack of prototypes, how about this:
> For building flex's own scanner, just forward declare the `rpl_malloc` and `rpl_realloc` prototypes despite whether we would be using them or not.
> This will fix the error. The long term solution would be to remove reliance on glibc's malloc behavior, but on standard one, however that would require more testing then.
> 
> -- 
> https://github.com/westes/flex/pull/248#issuecomment-326818782

-- 
Will Estes
westes575@gmail.com

hobbitalastair:
I've fixed the `make distcheck`failure (I hadn't used `$(srcdir)/` when referencing `scan.template`) - hopefully everything works properly now.

Explorer09:
@hobbitalastair Would you please see #261 ?

westes:
Thanks for your work on this. I went with #261 as an easier method, however.

PR #249, opened 2017-08-07
runtime for functional style code.

viktor-shepel:
**Issue:**
it is hard to read/analyze/modify code with deeply nested loops.

**Root cause:**
C language does not provide constructs for declarative computations.

**Solution:**
minimalistic FP runtime is introduced, small bit of code has loop
replaced with FP expressions as a showcase.

**Implementatin details:**
  * some operations defined recursively as gcc/clang compilers
    replace it with plain iterations under `-O2` flag which is
    utilized during build, see `-foptimize-sibling-calls` flag.
  * garbage collector justs track allocation and had to be
    launched manually, it will be replaced with conservative
    GC once whole FP idea will be accepted.

issue ticket - https://github.com/westes/flex/issues/250

### DIFF
--- a/src/functional/list.c
+++ b/src/functional/list.c
@@ -1,0 +1,53 @@
+#include "list.h"
+#include "gc.h"
+
+list_t* cons(void* head, void* tail) {
+	list_t* list = gc_malloc(sizeof *list);
+
+	return list ? (*list = (list_t) { head, tail }, list) : NULL;
+}
+
+void* head(list_t* list) {
+	return list->head; 
+}
+
+void* tail(list_t* list) {
+	return list->tail; 
+}
+
+bool is_empty(list_t* list) {
+	return list == NULL;
+}
+
+list_t* concat(list_t* left, list_t* right) {
+	return is_empty(left)
+		? right
+		: cons(head(left), concat(tail(left), right));
+}
+
+void* nth(size_t index, list_t* list) {
+	return index == 0 ? head(list) : nth(index - 1, tail(list));
+}
+
+size_t length(list_t* list) {
+	return is_empty(list) ? 0 : 1 + length(tail(list));
+}
+
+list_t* list_of_impl(size_t element_size, void* first, void* last) {
+	list_t* head = NULL;
+	char* start = (char*) first - element_size;
+	char* stop = (char*) last - element_size;
+	char* cursor = stop;
+
+	for (; cursor != start; cursor -= element_size) {
+		head = cons(cursor, head);
+	}
+
+	return head;
+}
+
+list_t* list_from(void* elements[], size_t size) {
+	return size == 0
+		? NULL
+		: cons(elements[0], list_from(&elements[1], size - 1));
+}
--- a/src/functional/list.h
+++ b/src/functional/list.h
@@ -1,0 +1,31 @@
+#ifndef LIST_H
+#define LIST_H
+
+#include <stddef.h>
+#include <stdbool.h>
+
+typedef struct list_t list_t;
+
+struct list_t {
+	void* head;
+	void* tail;
+};
+
+list_t* cons(void* head, void* tail);
+void* head(list_t* list);
+void* tail(list_t* list);
+bool is_empty(list_t* list);
+list_t* concat(list_t* left, list_t* right);
+void* nth(size_t index, list_t* list);
+size_t length(list_t* list);
+list_t* list_of_impl(size_t element_size, void* first, void* last);
+list_t* list_from(void* elements[], size_t size);
+
+#define list_of(type, first, last)							\
+	list_of_impl(sizeof(type), first, last)
+
+#define list(...) list_from(								\
+	(void* []) { __VA_ARGS__ },								\
+	sizeof (void* []) { __VA_ARGS__ } / sizeof (void*) )
+
+#endif
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -49,6 +49,18 @@ nodist_flex_SOURCES = \
 flex_CFLAGS = $(AM_CFLAGS) $(WARNINGFLAGS)
 
 COMMON_SOURCES = \
+	functional/boxes.c \
+	functional/boxes.h \
+	functional/character-class.c \
+	functional/character-class.h \
+	functional/function.c \
+	functional/function.h \
+	functional/gc.c \
+	functional/gc.h \
+	functional/list.c \
+	functional/list.h \
+	functional/operations.c \
+	functional/operations.h \
 	buf.c \
 	ccl.c \
 	dfa.c \
@@ -130,6 +142,18 @@ flex-stage1scan.$(OBJEXT): parse.h
 #   4. Run indent twice per file. The first time is a test.
 #      Otherwise, indent overwrites your file even if it fails!
 indentfiles = \
+	functional/boxes.c \
+	functional/boxes.h \
+	functional/character-class.c \
+	functional/character-class.h \
+	functional/function.c \
+	functional/function.h \
+	functional/gc.c \
+	functional/gc.h \
+	functional/list.c \
+	functional/list.h \
+	functional/operations.c \
+	functional/operations.h \
 	buf.c \
 	ccl.c \
 	dfa.c \
--- a/src/ecs.c
+++ b/src/ecs.c
@@ -33,37 +33,34 @@
 
 
 #include "flexdef.h"
+#include "functional/operations.h"
+#include "functional/character-class.h"
+
+static void deep_copy(list_t* destination, list_t* source) {
+	if (is_empty(source)) {
+		return;
+	}
+
+	deep_copy_character_class(head(destination), head(source));
+	deep_copy(tail(destination), tail(source));
+}
 
 /* ccl2ecl - convert character classes to set of equivalence classes */
 
 void    ccl2ecl (void)
 {
-	int     i, ich, newlen, cclp, ccls, cclmec;
+	list_t* character_classes = map(
+		character_class_at_index,
+		range(1, lastccl + 1)
+	);
 
-	for (i = 1; i <= lastccl; ++i) {
-		/* We loop through each character class, and for each character
-		 * in the class, add the character's equivalence class to the
-		 * new "character" class we are creating.  Thus when we are all
-		 * done, character classes will really consist of collections
-		 * of equivalence classes
-		 */
+	list_t* equivalence_classes = map(
+		to_equivalence_class,
+		character_classes
+	);
 
-		newlen = 0;
-		cclp = cclmap[i];
-
-		for (ccls = 0; ccls < ccllen[i]; ++ccls) {
-			ich = ccltbl[cclp + ccls];
-			cclmec = ecgroup[ich];
-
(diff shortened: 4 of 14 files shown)


Explorer09:
I wonder... How will this improve performance, or is it just change to the programming paradigm?
The code you introduced is much, much more complex than the original one.

Explorer09:
I have one more complaint though, you are defining new typedefs with the `_t` suffix naming. AFAIK, the `_t` suffix is reserved by POSIX. I will personally tolerate this for projects that keep watch on POSIX specs closely (for example the kernel, C library or core utilities), but I think it's better to avoid that in Flex.

src/functional/list.h
```diff
@@ -0,0 +1,31 @@
...
+#define LIST_H
+
+#include <stddef.h>
+#include <stdbool.h>
+
+typedef struct list_t list_t;
+
+struct list_t {
+	void* head;
+	void* tail;
```
Explorer09: Your definition of list_t is against the naming convention of linked lists.
Seriously I don't like it. Can we have a more conventional naming like this?

    struct linked_list_node {
    void *element;
    struct linked_list_node *next;
    }

src/functional/list.c
```diff
@@ -0,0 +1,53 @@
...
+}
+
+list_t* concat(list_t* left, list_t* right) {
+	return is_empty(left)
+		? right
+		: cons(head(left), concat(tail(left), right));
+}
+
+void* nth(size_t index, list_t* list) {
+	return index == 0 ? head(list) : nth(index - 1, tail(list));
```
Explorer09: Although tail call optimization is built-in for functional programming languages, it's not a good idea to assume a C compiler supports it. For example, tail call optimization is not performed on a debug build (with "no optimization" flag `cc -O0`), and if you not careful on the recursion limit, it can cause program to stack overflow and crash (for a network app this may be a vulnerability to Denial-of-Service attack). Did you notice this when having functional programming in C?

And one more thing:

    nth(index - 1, tail(list))

`tail(list)` is of type `void *`, and you implicitly cast it to `list_t *`. Be careful this casting may be invalid, resulting in invalid pointers when the list_t is processed further on. I don't think your style can even get accepted in C++ programs.

westes:
Would you link to where _t is reserved for POSIX?

On Monday, 14 August 2017,  4:52 am +0000, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> I have one more complaint though, you are defining new typedefs with the `_t` suffix naming. AFAIK, the `_t` suffix is reserved by POSIX. I will personally tolerate this for projects that keep watch on POSIX specs closely (for example the kernel, C library or core utilities), but I think it's better to avoid that in Flex.
> 
> -- 
> https://github.com/westes/flex/pull/249#issuecomment-322100020

-- 
Will Estes
westes575@gmail.com

Explorer09:
On Mon, Aug 14, 2017 at 6:37 PM, Will Estes <notifications@github.com> wrote:
>
> Would you link to where _t is reserved for POSIX?
>
http://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_02_02

In section 2.2.2: The Name Space, note the table row
Header: [CX] ANY header
Suffix: _t

viktor-shepel:
hi.
@Explorer09 thanks for feedback

> I wonder... How will this improve performance, or is it just change to the programming paradigm?
> The code you introduced is much, much more complex than the original one.

This pull request exposes sound abstractions (with help of programming paradigm) that should help fresh contributors grasp domain better, see explanation below.

Take for example `ccl2ecl` function in `src/ecs.c` file.
Essentially it converts `character classes`<sub>domain term</sub> to `equivalence classes`<sub>domain term</sub>.
Unfortunately there are no such entities in code and instead you had resort to 3 different arrays to express domain object, see diagram below. 
Operation on `character classes` requires juggling 3 arrays `cclmap/ccllen/ccltbl` in your head which is cumbersome<sub> actually it might be ok when performance is a goal but here we modify function that called once</sub>.
```c
/* representation of character class */

                 ccllen

                  +---+
                  | 2 +--------------------------------------+
                  +---+                                      |
                  | 1 +-------------------------+            |
                  +---+                         |            |
              +-->+ 0 +----------+              |            |
              |   +---+          |              |            |
              |                  v              v            v
              |     ccltbl     9 bytes       14 bytes      6 bytes
character     |            +-----------+----------------+---------+
  class   +---+            | f?c{1, 8} | [\x41|\x45]{3} | [ |~]|  |
  index       |            +-+---------+-+--------------+-+-------+
              |              ^           ^                ^
              |  cclmap      |           |                |
              |              |           |                |
              |   +---+      |           |                |
              +-->+ 0 +------+           |                |
                  +---+                  |                |
                  | 1 +------------------+                |
                  +---+                                   |
                  | 2 +-----------------------------------+
                  +---+


```
On other hand once we got domain object in code operation `ccl2ecl` becomes trivial:
```c
/* here we just transform character classes to equivalence one */
list_t* equivalence_classes = map(to_equivalence_class, character_classes);  
```

This pull request  does not degrade performance (though I need come up with use case where `ccl2ecl` invoked frequently to be more sure on statement).

quick comparison of execution times:
```
git checkout viktor-shepel-functional-style
./configure
make
perf stat -r 1000 -d src/flex examples/manual/pascal.lex

         18.435981 task-clock (msec)         #    1.328 CPUs utilized  ( +-  0.27% )
                30 context-switches          #    0.002 M/sec          ( +-  0.69% )
                 7 cpu-migrations            #    0.387 K/sec          ( +-  0.76% )
               899 page-faults               #    0.049 M/sec          ( +-  0.00% )
        10,437,298 cycles                    #    0.566 GHz            ( +-  2.46% )    

       0.013884943 seconds time elapsed                                ( +-  0.69% )

// VS

git checkout 535524ea6b55f8f6a7dfc160cb19211c084795d6  <-- one of the master commits
./configure
make
perf stat -r 1000 -d src/flex examples/manual/pascal.lex

         18.256896 task-clock (msec)         #    1.355 CPUs utilized  ( +-  0.27% )
                30 context-switches          #    0.002 M/sec          ( +-  0.64% )
                 7 cpu-migrations            #    0.399 K/sec          ( +-  0.69% )
               887 page-faults               #    0.049 M/sec          ( +-  0.00% )
        10,380,408 cycles                    #    0.569 GHz            ( +-  2.46% )

       0.013478181 seconds time elapsed                                ( +-  0.58% )
```

src/functional/list.h
viktor-shepel: sure thing.
Due to recursive definition there is no distinction between list and its node i.e. created node is a list by itself.

take for example Haskell definition
```hs
data LinkedList a = Empty | Cons a (LinkedList a) 
{- here LinkedList is our type and Empty and Cons are value constructors -}
```
analogy in C would be
```c
struct linked_list {
  void* head;
  struct linked_list* tail;
};

/* where Empty and Cons constructors replaced by NULL and cons respectively */
```

src/functional/list.c
viktor-shepel: if we support only prominent compilers `gcc/clang` then I'll add `-foptimize-sibling-calls` compiler flag to `configure.ac` it will replace recursion with iterations. Otherwise I could do it for compiler - change recursion to loops.

Explorer09:
On Thu, Aug 17, 2017 at 7:19 PM, Viktor <notifications@github.com> wrote:
>
> This pull request exposes sound abstractions (with help of programming paradigm) that should help fresh contributors grasp domain better, see explanation below.

I'm not against abstractions or paradigm changes, but I will consider
over-abstraction an issue.
And programming style, too. Your style doesn't fit how a C programmer
will program it.
By the way, my opinion does not represent Flex's decision.

> This pull request does not degrade performance (though I need come up with use case where ccl2ecl invoked frequently to be more sure on statement).
>
> quick comparison of execution times:
>
> git checkout viktor-shepel-functional-style
> ./configure
> make
> perf stat -r 1000 -d src/flex examples/manual/pascal.lex
>
>          18.435981 task-clock (msec)         #    1.328 CPUs utilized  ( +-  0.27% )
>                 30 context-switches          #    0.002 M/sec          ( +-  0.69% )
>                  7 cpu-migrations            #    0.387 K/sec          ( +-  0.76% )
>                899 page-faults               #    0.049 M/sec          ( +-  0.00% )
>         10,437,298 cycles                    #    0.566 GHz            ( +-  2.46% )
>
>        0.013884943 seconds time elapsed                                ( +-  0.69% )
>
> // VS
>
> git checkout 535524ea6b55f8f6a7dfc160cb19211c084795d6  <-- one of the master commits
> ./configure
> make
> perf stat -r 1000 -d src/flex examples/manual/pascal.lex
>
>          18.256896 task-clock (msec)         #    1.355 CPUs utilized  ( +-  0.27% )
>                 30 context-switches          #    0.002 M/sec          ( +-  0.64% )
>                  7 cpu-migrations            #    0.399 K/sec          ( +-  0.69% )
>                887 page-faults               #    0.049 M/sec          ( +-  0.00% )
>         10,380,408 cycles                    #    0.569 GHz            ( +-  2.46% )
>
>        0.013478181 seconds time elapsed                                ( +-  0.58% )
>

It's 3% slower, unfortunately.

src/functional/list.h
Explorer09: Don't assume that I didn't know what a linked list is.
Yes, a list with a node removed is still a list.
But in C, the list is not named that way, otherwise a reference like `my_list->head` could be confusing on what it does.
Syntax like `tail(tail(tail(my_list)))` is more hellish.
Please use C way of referencing it, that is: `my_list->next->next->next`

westes:
At the root of this, the main issue I see is:

Flex is a C program at heart.

I'm unclear on how the problem noted in the original post -- deeply nested code -- is solved by the proposed solution -- introducing abstractions that require a paradigm shift to express.

The cost seems high in terms of changing part of flex's codebase to do nnon-idiomatic things against the benefit of...well, I'm not sure quite what the benefit is.

Your thoughts, please.

On Thursday, 17 August 2017, 12:01 pm +0000, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> 
> 
> 
> > @@ -0,0 +1,31 @@
> +#ifndef LIST_H
> +#define LIST_H
> +
> +#include <stddef.h>
> +#include <stdbool.h>
> +
> +typedef struct list_t list_t;
> +
> +struct list_t {
> +	void* head;
> +	void* tail;
> 
> Don't assume that I didn't know what a linked list is.
> Yes, a list with a node removed is still a list.
> But in C, the list is not named that way, otherwise a reference like `my_list->head` could be confusing on what it does.
> Syntax like `tail(tail(tail(my_list)))` is more hellish.
> Please use C way of referencing it, that is: `my_list->next->next->next`
> 
> -- 
> https://github.com/westes/flex/pull/249#discussion_r133692991

-- 
Will Estes
westes575@gmail.com

viktor-shepel:
it actually may be slower.
note that perf is a sampling profiler plus there are different сonfidence interval 0.69 VS 0.58.

viktor-shepel:
@westes 
lets drop this PR/issue then

westes:
Closing due to issues discussed in comments.

Issue #251, opened 2017-08-16
#line 0 is considered an error by HP'-UX's ANSI C compiler


rklrkl:
With flex 2.6.4, the files src/main.c (line 495) and src/misc.c (line 369) both produce "#line 0" output in src/stage1scan.c when stage1flex is run during a make (ENABLE_BOOTSTRAP_TRUE code in src/Makefile.in at line 1696).

Unfortunately., HP-UX 11.31's HP ANSI C compiler considers - probably correctly! - that "#line 0" is an illegal line number and refuses to the compile the generated C file. I changed src/main.c and src/misc.c to use the value 1 instead of 0 and it happily compiled the lines now containing "#line 1".

I would suggest a similar change to the flex source to remedy this issue.

westes:
Please submit a pull request; it makes reviewing your proposed change and attributing it to you easier.

On Wednesday, 16 August 2017,  1:23 pm +0000, rklrkl <notifications@github.com> wrote:

> With flex 2.6.4, the files src/main.c (line 495) and src/misc.c (line 369) both produce "#line 0" output in src/stage1scan.c when stage1flex is run during a make (ENABLE_BOOTSTRAP_TRUE code in src/Makefile.in at line 1696).
> 
> Unfortunately., HP-UX 11.31's HP ANSI C compiler considers - probably correctly! - that "#line 0" is an illegal line number and refuses to the compile the generated C file. I changed src/main.c and src/misc.c to use the value 1 instead of 0 and it happily compiled the lines now containing "#line 1".
> 
> I would suggest a similar change to the flex source to remedy this issue.
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/westes/flex/issues/251

-- 
Will Estes
westes575@gmail.com

Explorer09:
@rklrkl Could you please paste us a diff between the src/stage1scan.c that your machine generates and src/scan.c in flex-2.6.4 release tarball?
I wonder what happened to the line number generation.
This may not be the fault of the compiler, but something is wrong with your first stage Flex's code generation. (Ideally for release version the src/stage1scan.c and src/scan.c should be almost identical except the filenames, unless you specify other LFLAGS during the bootstrap.)

Issue #235, opened 2017-07-01
[2.6.4] #line number distorted if indented initial statements between first %% and first rule


jannick0:
The #line numbers are distorted if (indented) statements between the first %% and the first rule are present. Commit 2c7e34bb958ac5d083b2bab4c89907cd0669196d might be linked to this issue.

Example:
```c
%option noyywrap
%%
    printf("START\n"); /// this line distorts the #line numbers; issue disappears after removing this line (compare .yy.c)
.
%%
int main(int argc, char** argv)
{
	return 0;
}`

westes:
Ah, looking at the AUR page you link to, there is a solution there. Note that the solution has nothing to do with flex. Closing this bug as invalid.

westes:
Ack, I commented on the wrong issue...sigh.

westes:
Does this happen against master?

jannick0:
I would love to build the trunk version myself, but that fails on my Win10 box togehter with cygwin with the error report below. Apparently a gettext-tools version confusion since I am ahead of version 18.0 with 0.19.8.1. 

Separate question: Is there any way (e.g cmake) to avoid the libtoolize route for building flex?

```bash
$ ./autogen.sh
libtoolize (GNU libtool) 2.4.6
Written by Gary V. Vaughan <gary@gnu.org>, 2003

Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'build-aux'.
libtoolize: linking file 'build-aux/config.guess'
libtoolize: linking file 'build-aux/config.sub'
libtoolize: linking file 'build-aux/install-sh'
libtoolize: linking file 'build-aux/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: linking file 'm4/libtool.m4'
libtoolize: linking file 'm4/ltoptions.m4'
libtoolize: linking file 'm4/ltsugar.m4'
libtoolize: linking file 'm4/ltversion.m4'
libtoolize: linking file 'm4/lt~obsolete.m4'
FIND: Parameterformat falsch
autopoint: *** infrastructure files for version 0.18 not found; this is autopoint from GNU gettext-tools 0.19.8.1
autopoint: *** Stop.
[...]

PR #252, opened 2017-08-23
Line directive correction package

jannick0:
Here a sequence of commits covering the `#line` directive issues (#235 and, potentially, #251) which leave the code created by flex untouched. I hope the changes are well chopped to pieces to make the changes understandable.

Separately - and in case someone wants to compare two versions of `flex` - attached a small [zip package](https://github.com/westes/flex/files/1247262/line-dir_2017-08-24.zip) for a `#line` directive consistency check and a code text regression test (based on a `make` file invoking two `awk` scripts). Note that you need `make`, `gawk`, `sed` and `diff` to run the tests. Both of them run against the test `.l` files in the `tests` folder show that the commits do not change the code generated by `flex` as such.

Some summarizing notes:

* the line directives created by flex are now consistent for `.l` files and can be used for debugging; I partially rolled back commit  4bffa41 for code transparency reasons. It took me awfully long to check things before I decided to eliminate the macros introduced by this commit, whereas after making them explicit again it was kind of easy to see what is going on ... and wrong. I hope this is fine with the author of the mentioned commit.
* the case statements in the function `yylex` are now closed by `#line` directives relative to the flex generated code (in line with `bison`'s output)
* cleaning up the functions `buf_linedir` and `filter_fix_linedirs`; the latter is now independent of `<regex.h>` after Explorer09 took the first step in that direction in commit e784a80.  The full removal of this dependency is not contained in the package, but should be easy I believe. However, I must admit that I better keep away from the `autotools` monster.
* various bugs fixed and some minor edits applied 

Hope this is useful.

### DIFF
--- a/src/buf.c
+++ b/src/buf.c
@@ -95,7 +95,7 @@ struct Buf *buf_linedir (struct Buf *buf, const char* filename, int lineno)
     const char *src;
     size_t tsz;
 
-    if (gen_line_dirs)
+    if (!gen_line_dirs)
 	return buf;
 
     tsz = strlen("#line \"\"\n")                +   /* constant parts */
--- a/src/buf.c
+++ b/src/buf.c
@@ -91,28 +91,13 @@ struct Buf *buf_prints (struct Buf *buf, const char *fmt, const char *s)
  */
 struct Buf *buf_linedir (struct Buf *buf, const char* filename, int lineno)
 {
-    char *dst, *t;
-    const char *src;
-    size_t tsz;
-
     if (!gen_line_dirs)
 	return buf;
 
-    tsz = strlen("#line \"\"\n")                +   /* constant parts */
-               2 * strlen (filename)            +   /* filename with possibly all backslashes escaped */
-               (size_t) (1 + ceil (log10 (abs (lineno)))) +   /* line number */
-               1;                                   /* NUL */
-    t = malloc(tsz);
-    if (!t)
-      flexfatal (_("Allocation of buffer for line directive failed"));
-    for (dst = t + snprintf (t, tsz, "#line %d \"", lineno), src = filename; *src; *dst++ = *src++)
-      if (*src == '\\')   /* escape backslashes */
-        *dst++ = '\\';
-    *dst++ = '"';
-    *dst++ = '\n';
-    *dst   = '\0';
-    buf = buf_strappend (buf, t);
-    free(t);
+    char directive[MAXLINE];
+    sprintf(directive, "#line %d \"%s\"\n", lineno, filename);
+    buf = buf_strappend(buf, directive);
+
     return buf;
 }
 
--- a/src/filter.c
+++ b/src/filter.c
@@ -22,6 +22,9 @@
 /*  PURPOSE. */
 
 #include "flexdef.h"
+
+#define M4_LINE_DIR_DUMMY_BEG      "#line XYZ "
+const char* m4_line_dir_dummy     = M4_LINE_DIR_DUMMY_BEG "\"M4_YY_OUTFILE_NAME\"\n";
 static const char * check_4_gnu_m4 =
     "m4_dnl ifdef(`__gnu__', ,"
     "`errprint(Flex requires GNU M4. Set the PATH or set the M4 environment variable to its path name.)"
@@ -293,7 +296,7 @@ int filter_tee_header (struct filter *chain)
 
 		/* write a fake line number. It will get fixed by the linedir filter. */
 		if (gen_line_dirs)
-			fprintf (to_h, "#line 4000 \"M4_YY_OUTFILE_NAME\"\n");
+			fprintf (to_h, m4_line_dir_dummy);
 
 		fprintf (to_h, "#undef %sIN_HEADER\n", prefix);
 		fprintf (to_h, "#endif /* %sHEADER_H */\n", prefix);
@@ -332,10 +335,11 @@ static bool is_blank_line (const char *str)
 }
 
 /** Adjust the line numbers in the #line directives of the generated scanner.
- * After the m4 expansion, the line numbers are incorrect since the m4 macros
- * can add or remove lines.  This only adjusts line numbers for generated code,
- * not user code. This also happens to be a good place to squeeze multiple
- * blank lines into a single blank line.
+ * After the m4 expansion, the line numbers might not be incorrect since the 
+ * m4 macros can add or remove lines.  This only fills in line numbers for 
+ * generated code (labelled as XYZ), not user code. 
+ * This also happens to be a good place to squeeze multiple blank lines into 
+ * a single blank line.
  */
 int filter_fix_linedirs (struct filter *chain)
 {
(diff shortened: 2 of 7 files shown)


westes:
I will review this. If your tests look good, I will incorporate them into the test suite.

Autotools is not so arcane as it looks. It's just that autotools was not written with the case of bootstrapping things like flex in mind so we have to abuse it a fair bit to make it do things that its instincts tell it should not be possible.

On Wednesday, 23 August 2017, 11:04 pm +0000, jannick0 <notifications@github.com> wrote:

> Here a sequence of commits covering the `#line` directive issues (#235 and, potentially, #251) which leave the code created by flex untouched. I hope the changes are well chopped to pieces to make the changes understandable.
> 
> Separately - and in case someone wants to compare two versions of `flex` - attached a small [zip package](https://github.com/westes/flex/files/1247262/line-dir_2017-08-24.zip) for a `#line` directive consistency check and a code text regression test (based on a `make` file invoking two `awk` scripts). Note that you need `make`, `gawk`, `sed` and `diff` to run the tests. Both of them run against the test `.l` files in the `tests` folder show that the commits do not change the code generated by `flex` as such.
> 
> Some summarizing notes:
> 
> * the line directives created by flex are now consistent for `.l` files and can be used for debugging; I partially rolled back commit  4bffa41 for code transparency reasons. It took me awfully long to check things before I decided to eliminate the macros introduced by this commit, whereas after making them explicit again it was kind of easy to see what is going on ... and wrong. I hope this is fine with the author of the mentioned commit.
> * the case statements in the function `yylex` are now closed by `#line` directives relative to the flex generated code (in line with `bison`'s output)
> * cleaning up the functions `buf_linedir` and `filter_fix_linedirs`; the latter is now independent of `<regex.h>` after Explorer09 took the first step in that direction in commit e784a80.  The full removal of this dependency is not contained in the package, but should be easy I believe. However, I must admit that I better keep away from the `autotools` monster.
> * various bugs fixed and some minor edits applied 
> 
> Hope this is useful.
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/westes/flex/pull/252
> 
> -- Commit Summary --
> 
>   * bug fix in scan.l - correcting emitted line directives by (1) removing
>   * bug fix - buf.c: corrected mode to generate line directives
>   * scan.l - convert backslashes in 'infilename' to forward slashes when
>   * buf.c - cleaning up of function 'buf_linedir' (given that filename does
>   * consistent use of '#line XYZ "M4_YY_OUTFILE_NAME"' as generic line
>   * add closing line directive (relative to flex output file)
> 
> -- File Changes --
> 
>     M src/buf.c (25)
>     M src/filter.c (76)
>     M src/flexdef.h (1)
>     M src/main.c (2)
>     M src/misc.c (51)
>     M src/parse.y (7)
>     M src/scan.l (53)
> 
> -- Patch Links --
> 
> https://github.com/westes/flex/pull/252.patch
> https://github.com/westes/flex/pull/252.diff
> 
> -- 
> https://github.com/westes/flex/pull/252

-- 
Will Estes
westes575@gmail.com

Explorer09:
A minor complaint: You should properly wrap the lines in the commit description, for example in your 00622bd8c75f66067e980d3be46ce1ed4e42c248 your newlines are all missing. That would make the commit unreadable in GUI.

And then in 8caa9e4936ebf6d094c409643a90df745527e90f:
I don't like the XYZ placeholder. If you're really serious, define a flex internal magic word is a better idea.
I give one suggestion: `__FLEX_LINENO_PLACEHOLDER`

Explorer09:
Here's one more issue in f8a7cebb8758584a7b67e4965c995096ef88f752:
I may need to warn you about a potential issue when converting the backslashes to slashes in filenames:
In some CJK encodings (notably Big5, GBK and Shift_JIS), there are Han character code points that contain 0x5C bytes. It'll bring problems if you convert them byte-by-byte ignoring the current system's encoding.
We people in Taiwan have a nickname for this problem: The Xu Gai Gong (許蓋功) problem.

You can ignore this if you use Flex only on UTF-8 files. But if Flex was intended to also work on other encodings, this would break them.

westes:
Flex is intended to work in any locale. While there are some issues with non-latin-* encodings, at a bare minimum, anything that flex does should respect the existing locale, or at least not break it.

On Thursday, 24 August 2017,  8:10 pm -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> Here's one more issue in f8a7cebb8758584a7b67e4965c995096ef88f752:
> I may need to warn you about a potential issue when converting the backslashes to slashes in filenames:
> In some CJK encodings (notably Big5, GBK and Shift_JIS), there are Han character code points that contain 0x5C bytes. It'll bring problems if you convert them byte-by-byte ignoring the current system's encoding.
> We people in Taiwan have a nickname for this problem: The Xu Gai Gong (許蓋功) problem.
> 
> You can ignore this if you use Flex only on UTF-8 files. But if Flex was intended to also work on other encodings, this would break them.
> 
> -- 
> https://github.com/westes/flex/pull/252#issuecomment-324812696

-- 
Will Estes
westes575@gmail.com

Explorer09:
And I forgot that embedded backslashes in filenames are allowed in Unix.

jannick0:
@Explorer09 :
> A minor complaint: You should properly wrap the lines in the commit description, for example in your 00622bd your newlines are all missing. That would make the commit unreadable in GUI.

As far as I am concerned, the line breaks are wonderfully displayed in the commit description on github, so I would conclude they do exist. Of course, I cannot say anything about other apps displaying these comments.

> And then in 8caa9e4:
> I don't like the XYZ placeholder. If you're really serious, define a flex internal magic word is a better idea.
> I give one suggestion: __FLEX_LINENO_PLACEHOLDER

That would be fine with me. Easy enough to simply change the name. In the light of the expected line directives flex recognizes (see paragraph below), this would be a minor issue.

> EDIT: I should mention that preprocessing token is allowed in #line directive in C99. So a syntax like #line FOO_BAR could be seen in a valid user code.

Correct, but your example is not parsed as such by flex, so no need to worry about this, unless a cpp parser should be implemented here, which is currently not the case. For now - and this is sufficient in my view - line directives of the form `#line number "..."` in flex input files are properly processed. (NB: What flex currently accepts as ... is very limited in contrast what the cpp standard is like.)

> Here's one more issue in f8a7ceb:
> I may need to warn you about a potential issue when converting the backslashes to slashes in filenames:
> In some CJK encodings (notably Big5, GBK and Shift_JIS), there are Han character code points that contain 0x5C bytes. It'll bring problems if you convert them byte-by-byte ignoring the current system's encoding.
> We people in Taiwan have a nickname for this problem: The Xu Gai Gong (許蓋功) problem (or 許功蓋 - because these three characters have the problem in Big5).
> 
> You can ignore this if you use Flex only on UTF-8 files. But if Flex was intended to also work on other encodings, this would break them.

Thanks, that is interesting. Given that I am not an expert at CJK encodings and alike, there are some questions then begging for answers where the most important one is how multi-byte encodings are (intrinsically) treated in c code to cope with this issue. Otherwise the bare objective that flex should not break anything is difficult to achieve I guess when it comes to line directives created by flex (in contrast to those line directives already existing in flex input).

Anyone with implementation suggestions?

I am hoping that the flex output could be properly debugged soon and - but this is another issue, of course - flex itself, as well. I was trying to contribute for the first goal, but Will and others, please do feel free to amend as you feel like and may find appropriate.

Explorer09:
@jannick0 

> As far as I am concerned, the line breaks are wonderfully displayed in the commit description on github, so I would conclude they do exist. Of course, I cannot say anything about other apps displaying these comments.

It might be just me, but I do read git commit descriptions through `git log` or emails, and so I expect commit descriptions be properly wrapped (to 72 chars per line - you need to leave space for `git log` indents).

Software doesn't wrap that for you due to concerns that some data / program outputs should be included as is without wrapping, and it will be you to distinguish the two.

These are just "good practices" of making git patches, and I won't repeat what textbooks have said.

> Correct, but your example is not parsed as such by flex, so no need to worry about this, unless a cpp parser should be implemented here, which is currently not the case. For now - and this is sufficient in my view - line directives of the form `#line number "..."` in flex input files are properly processed. (NB: What flex currently accepts as ... is very limited in contrast what the cpp standard is like.)

Actually I'm not expecting Flex will support full pre-processing. I was expecting #line directives with pre-processing tokens be ignored by Flex and copied as is to the generated file. What I concern is the namespace pollution just because of developer's oversight. `#line XYZ` should be user's code, not something that Flex would generate, even in the middle phase.

westes:
Yes, git commit messages should be wrapped at 72 characters. I have been known to (silently) do this to fix things, but the less work you all make for me, the better. I will get around to adding hooks to enforce this at some point, but putting infrastructure to handle releases more better is higher priority -- unless one of you wants to assist with a pull request.

On Tuesday, 12 December 2017,  9:40 am -0800, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> @jannick0 
> 
> > As far as I am concerned, the line breaks are wonderfully displayed in the commit description on github, so I would conclude they do exist. Of course, I cannot say anything about other apps displaying these comments.
> 
> It might be just me, but I do read git commit descriptions through `git log` or emails, and so I expect commit descriptions be properly wrapped (to 72 chars per line - you need to leave space for `git log` indents).
> 
> Software doesn't wrap that for you due to concerns that some data / program outputs should be included as is without wrapping, and it will be you to distinguish the two.
> 
> These are just "good practices" of making git patches, and I won't repeat what textbooks have said.
> 
> > Correct, but your example is not parsed as such by flex, so no need to worry about this, unless a cpp parser should be implemented here, which is currently not the case. For now - and this is sufficient in my view - line directives of the form `#line number "..."` in flex input files are properly processed. (NB: What flex currently accepts as ... is very limited in contrast what the cpp standard is like.)
> 
> Actually I'm not expecting Flex will support full pre-processing. I was expecting #line directives with pre-processing tokens be ignored by Flex and copied as is to the generated file. What I concern is the namespace pollution just because of developer's oversight. `#line XYZ` should be user's code, not something that Flex would generate, even in the middle phase.
> 
> -- 
> https://github.com/westes/flex/pull/252#issuecomment-351127569

-- 
Will Estes
westes575@gmail.com

Issue #253, opened 2017-08-24
Provide a pkg-config for libfl


Slabity:
It would be nice to be able to use pkg-config to link to libfl.

westes:
Submit a pull request, then. Or at a bare minimum, link me to where it is described how to set that sort of thing up.

Slabity:
Wow, that was a quick response.

Sorry, I'm working on the pc file now.

PR #254, opened 2017-08-24
Create libfl.pc target for pkgconfig

Slabity:
This adds a target to generate `libfl.pc`, a file used by `pkg-config` to locate shared objects and header files.


Slabity:
Adds functionality requested in #253

westes:
I am making the following changes to this pull request:

* In Makefile.am, I moved the block describing the new file libfl.pc to earlier. This is for reading convenience.
* In .gitignore, I added libfl.pc so that git will not pay attention to that file since it is generated by running configure.

westes:
Thanks, this is now on master and will be included in the next release of flex.

PR #257, opened 2017-08-26
libfl.pc changes

Explorer09:
Follow-up of pull request #254.
Three minor changes on libfl.pc - see commit description.

PR #258, opened 2017-08-28
Convenience makefile targets to build libfl only

Explorer09:
See #256 for the discussion.
Note that I base this pull request on #257 so that we can have libfl.pc in the `src` subdirectory instead of top directory.


westes:
In several places, you have constructions that begin "cd src", but shouldn't those be "cd $(top_srcdir)/src" or "cd $(top_builddir)/src" -- depending on which instance we're talking about?

Explorer09:
@westes It's actually "cd $(builddir)/src" (but $(builddir) always equal to "." for autoconf).
The purpose is only to invoke src/Makefile with the specified targets. The makefiles generated are always in top_builddir, so `cd src && make ...` is sufficient.

westes:
Thanks. this is on master and will be included in the next release.

PR #260, opened 2017-09-03
Forward declare rpl_* functions.

### LINKED ISSUE #247: rpl_* functions and config.h/stdlib.h interations causes segfault with modern GCC

Explorer09:
Better than #248 approach.
Fixes #247.
Note to @hobbitalastair: I didn't get a segfault when I test bootstrapping this, but at least I killed compiler warnings.

### DIFF
--- a/src/flexdef.h
+++ b/src/flexdef.h
@@ -37,6 +37,12 @@
 
 #ifdef HAVE_CONFIG_H
 #include <config.h>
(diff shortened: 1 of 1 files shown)


westes:
@Explorer09 why did you close this?

Explorer09:
@westes Because I think my fix failed to address another problem that would happen besides #247. Please wait for another pull request.

PR #262, opened 2017-09-04
AC_USE_SYSTEM_EXTENSIONS in configure.ac

### LINKED ISSUE #241: [2.6.4 clang/glibc-2.25.90] implicit declaration of function 'reallocarray' is invalid in C99
When built with clang-4.0.1, on system with glibc-2.25.90 (glibc 2.26 development snapshot, current git master), the following is noted during compilation

https://gist.github.com/elkrejzi/baad3a9c0035ef954cbb8e6ce2d635ba

At runtime, while compiling doxygen, flex will crash with the following backtrace

https://gist.github.com/elkrejzi/2f821eab0346ea1d1501dfd0f8c53195

Explorer09:
This would define _GNU_SOURCE in config.h, enabling the reallocarray()
prototype in glibc 2.26+.

This would really fix #241.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -25,8 +25,10 @@
 # autoconf requirements and initialization
 
 AC_INIT([the fast lexical analyser generator],[2.6.4],[flex-help@lists.sourceforge.net],[flex])
+AC_PREREQ([2.60])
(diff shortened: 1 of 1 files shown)


westes:

PR #266, opened 2017-09-22
Fix HAVE_REALLOCARRAY check

lukbaj:


Explorer09:
I think it's better to remove that declaration altogether. We longer define our reallocarray() function, so just use the prototype from <stdlib.h> is enough.

lukbaj:
Done

westes:

PR #267, opened 2017-09-24
Simplify PRINT_SPACES() macro

Explorer09:
A trivial optimization for size :)
(By the way, scanopt_usage is currently an unused function in flex.)


westes:

Issue #263, opened 2017-09-04
flex.skl calls fileno(), which is not ISO C99 function


Explorer09:
I caught this warning when building flex in `-std=c99` mode:

```
./stage1flex   -o stage1scan.c ./scan.l
  CC       flex-stage1scan.o
stage1scan.c: In function ‘yy_init_buffer’:
stage1scan.c:4757:48: warning: implicit declaration of function ‘fileno’; did you mean ‘mblen’? [-Wimplicit-function-declaration]
         b->yy_is_interactive = file ? (isatty( fileno(file) ) > 0) : 0;
                                                ^~~~~~
                                                mblen
```

In glibc, fileno() prototype is exposed only if `_POSIX_C_SOURCE` is defined to `>= 1`. Although flex will build with `_GNU_SOURCE` defined (AC_USE_SYSTEM_EXTENSIONS in configure script), the problem is that generated scanners may not build if the user's environment requires strict ISO C99, and I consider this a bug.

So this issue report is about two things:
1. The `#define _POSIX_C_SOURCE 1` is missing in flex.skl when the scanner is specified neither `always-interactive` or `never-interactive` option.
2. We didn't document that we use POSIX functions to detect whether the input is "interactive", and so brings surprise to users that expect it builds in non-POSIX environments.

westes:
What about rewriting to not use fileno() and being C99 compliant that way?

westes:
I checked against my copy of the glibc reference manual and it doesn't note anything special about fileno() as well. Which is unfortunate given what you describe.

Explorer09:
@westes 
For your first question, it won't help much if we replace fileno() with something else since the routine already depends on isatty(), and so I'm pursuing only to have the feature macro defined.
For your second question, glibc manual doesn't explicitly mention about feature macro requirements, but Linux man pages does, as well as the code of "/usr/include/stdio.h" and "features.h":

```
/*stdio.h*/
#ifdef	__USE_POSIX
/* Return the system file descriptor for STREAM.  */
extern int fileno (FILE *__stream) __THROW __wur;
#endif /* Use POSIX.  */
```

```
/*features.h*/
#if defined _POSIX_SOURCE || _POSIX_C_SOURCE >= 1 || defined _XOPEN_SOURCE
# define __USE_POSIX	1
#endif
```

westes:
On Monday,  4 September 2017,  8:44 am -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

[...]

PR #269, opened 2017-10-08
Define _POSIX_C_SOURCE when needed in skeleton.

Explorer09:
This is a quick fix of #263 where we lack feature macro to expose the API of fileno().


westes:

PR #270, opened 2017-10-12
Temporary protection against ccl overflow & overwriting

Explorer09:
For ccladd(), if cclp given is a non-last ccl, adding a char into it
will overflow the buffer and overwrite the first char in the next ccl.

For now, add a temporary detection and protection code.
(Not sure if this could happen in user input, but if it could, then
you can expect some "corrupted" behavior for generated scanners.)


westes:

PR #272, opened 2017-10-13
Fix add_action() undefined behavior (int overflow).

Explorer09:


westes:

PR #273, opened 2017-10-13
Remove unused BASENAME() macro

Explorer09:
(Simply forgot to do so in commit fb731ac0221e1866534dfe072b84b8af7a5d88f3)


westes:
Thanks. THis is on master and will be included in the next release of flex.

PR #278, opened 2017-10-27
Fixed twice assigned value

lunixoid:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. A warnings, found using PVS-Studio:
flex/src/scanopt.c  154     warn    V519 The 's->optc' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 144, 154.

### DIFF
--- a/src/scanopt.c
+++ b/src/scanopt.c
@@ -141,7 +141,6 @@ scanopt_t *scanopt_init (const optspec_t *options, int argc, char **argv, int fl
 	s = malloc(sizeof (struct _scanopt_t));
 
 	s->options = options;
-	s->optc = 0;
 	s->argc = argc;
 	s->argv = (char **) argv;
 	s->index = 1;


Explorer09:
> Also static analyser found warnings in scanopt.c (wasn't fixed in my PR):
> flex/src/scanopt.c	143	warn	V522 There might be dereferencing of a potential null pointer 's'. Check lines: 143, 141.

How about marking the scanopt_init function with `__attribute__((nonnull(1)))` so that the warning could be suppressed for now?
This function is currently only called once, in flexinit() in main.c. I think the whole scanopt thing could be replaced with getopt_long() in the long term, avoiding reinventing the wheel.

lunixoid:
I will test it little bit later.

lunixoid:
@Explorer09 The same entry:
flex/src/scanopt.c	143	warn	V522 There might be dereferencing of a potential null pointer 's'. Check lines: 143, 141.

Explorer09:
@lunixoid
> The same entry:
> flex/src/scanopt.c	143	warn	V522 There might be dereferencing of a potential null pointer 's'. Check lines: 143, 141.

Then I think you should ignore the warning then. It's too bad that your code analyzer can't catch the attribute syntax and suppress the warning. Many routines simply assert that the pointer arguments passed into will never be NULL, and so they can save a few instructions doing the runtime check. It's easy to find such examples in standard library, here's one: `memcpy()`

lunixoid:
Ok.

PR #279, opened 2017-10-30
Obsolete yypad64() macro.

Explorer09:
Slightly rewrite the logic in yytbl_data_load() and yytbl_write_pad64()
so they simply check if the bytes read/written are in 8-byte boundary.
No need to calculate how many bytes we need to pad. (Incidentally this
makes smaller code in x86_64.)

For yytbl_hdr_init(), just expand the calculation from the macro.


westes:

PR #282, opened 2017-11-02
Consistently use #ifdef for ENABLE_NLS check

arichardson:
config.h will have either define ENABLE_NLS or not define it. If it is not
defined we get a -Wundef warning due to using #if with an undefined macro

### DIFF
--- a/src/gettext.h
+++ b/src/gettext.h
@@ -20,7 +20,7 @@
 #define _LIBGETTEXT_H 1
 
 /* NLS can be disabled through the configure --disable-nls option.  */
-#if ENABLE_NLS
+#ifdef ENABLE_NLS
 
 /* Get declarations of GNU message catalog functions.  */
 # include <libintl.h>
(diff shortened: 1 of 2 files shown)


westes:

Explorer09:
Wait!
@arichardson, why aren't you also address the HAVE_LOCALE_H macro right below in main.c the same way?
Also, I don't think you should modify gettext.h - it's from the gettext package, and we should sync from there.
Pardon to ask, what exact compile warnings do you get during compiling gettext.h?

arichardson:
I was compiling using clang and -pedantic flag. I assume I didn't get a [-Wundef](https://clang.llvm.org/docs/DiagnosticsReference.html#wundef) warnuing for HAVE_LOCALE_H either because it gets defined to 0/1 or it was defined for me.

Explorer09:
@arichardson You won't get the warning for HAVE_LOCALE_H because it's compiled out after ENABLE_NLS is evaluated.
And why are you compiling with -Wundef option? AFAIK, it's an optional warning, and C99 already requires that undefined symbols be treated as 0 during #if evaluations. The warning option is only useful for pre-C99 compatibility.

PR #284, opened 2017-11-06
build: proper fix for ENABLE_NLS preprocessor check

Explorer09:
Because ENABLE_NLS may be defined to 0 (manually, not through autoconf)
and it's semantically incorrect to only check whether it's defined.

This is a correction to commit 661d603b65385f62f372acd2017e5af2e0f0cd50.


westes:

PR #286, opened 2017-11-12
Allow "--" as option argument.

### LINKED ISSUE #285: "--" should be allowed as option argument
A bug discovered in scanopt() parsing logic. However it's easy to fix.

Assume wc1.l is a valid flex file (I take this from "example/fastwc/wc1.l" in flex source).

Then, this will succeed and generate an output file named "--":
```
$ flex --outfile=-- wc1.l
```

But this would fail with message "flex: option `--outfile' requires an argument":
```
$ flex --outfile -- wc1.l
```

You can compare this with how GNU sed interprets option argument "--":
```
$ echo 'p' >./--
$ sed -n --file -- wc1.l
```

So flex rejecting "--" as option argument is a bug.

Explorer09:
Fixes #285, where invocation like this would fail with "requires an
argument" error message:

    flex --outfile -- wc1.l

### DIFF
--- a/src/scanopt.c
+++ b/src/scanopt.c
@@ -733,8 +733,7 @@ int     scanopt (scanopt_t *svoid, char **arg, int *optindex)
 
 	/* Look ahead in argv[] to see if there is something
 	 * that we can use as an argument (if needed). */
-	has_next = s->index + 1 < s->argc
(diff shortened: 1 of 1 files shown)


westes:

Issue #290, opened 2017-12-09
yy_set_interactive is undefined


rmccullagh:
Running `master` and trying to use yy_set_interactive(1); and I receive a undefined reference when compiling on Ubuntu 16.04

Here is my .l file options:

%option warn nodefault
%option reentrant never-interactive 
%option bison-bridge bison-locations yylineno
%option nounput noinput noyywrap unistd

Explorer09:
Go ahead and file a pull request on GitHub. That way we can review what changes you will propose.

Issue #277, opened 2017-10-27
skeleton: yywrap() forward declaration doesn't allow static definition


Explorer09:
This lack of feature brings a small hindrance in code optimization.
I discovered this when I tried to redefine Flex's own scanner's yywrap() function to static:

```diff
diff --git a/src/scan.l b/src/scan.l
index 5ac864a..8866c33 100644
--- a/src/scan.l
+++ b/src/scan.l
@@ -1019,7 +1019,7 @@
 %%
 
 
-int yywrap(void)
+static int yywrap(void)
        {
        if ( --num_input_files > 0 )
                {
```
And it broke with this gcc error:
```
scan.l:1022:12: error: static declaration of ‘yywrap’ follows non-static declaration
```
It doesn't seem like flex's own yywrap() function is called from outside `scan.l` and so ideally should be declared static. But flex's skeleton and generator doesn't yet allow this static definition.

jannick0:
Would the suggestion imply that every `yywrap` function in any `flex` output - apart of the 2-stage-compilation approach of `flex` - has to be `static`, too? In this case this would most probably break existing code using `flex`, hence I would opt not to support this. 

I understand the suggestion as an enhancement of the `flex` code for which the following could be considered:

- There is already the macro `YY_SKIP_WRAP` to overwrite the standard declaration of `yywrap`. 
- If you want to handle different versions of `yywrap` functions for each scanner, then you could simply use the `prefix` feature and add a different prefix to each of the `yywrap` functions.

Explorer09:
@jannick0
You misunderstood about the static thing. I didn't say it _has to_ be static; I'm saying that it _could_ be static, and it's better have a way to specify so.
I don't think YY_SKIP_WRAP is the right macro for this, since it's a boolean definition, and the name won't stand what it's for if we adopt it for this use, anyway.
The `prefix` isn't either. It's supposed be a namespace prefix, you know? My feature request isn't about namespace, but to allow a custom declaration of `yywrap` so I can specify `static` or `inline` or gcc's `__attribute__`or whatever I want.

jannick0:
@Explorer09 :
> You misunderstood about the static thing. I didn't say it has to be static; I'm saying that it could be static, and it's better have a way to specify so.

I am still struggling to understand if this should apply to `yywrap` as part of either the flex package or flex output (c / c++) or even both. Regardless of having an option or not, the remainder of the flex code should be changed to cover the case when the option to define `yywrap` as `static` is on.

[...]

PR #291, opened 2017-12-12
implement yyset_interactive [#290]

rmccullagh:
Currently, only available in reentrant mode, since it needs a scanner argument. However, I believe a `yyinteractive` macro can be implemented too.


src/flex.skl
```diff
@@ -618,6 +619,7 @@ void yyfree ( void * M4_YY_PROTO_LAST_ARG );
 
 m4_ifdef( [[M4_YY_NOT_IN_HEADER]],
 [[
+
 #define yy_new_buffer yy_create_buffer
 #define yy_set_interactive(is_interactive) \
```
Explorer09: When you are implementing yyset_interactive, could you also remove this yy_set_interactive macro, which could cause confusion.

jannick0:
Sorry for my ignorance: Why not simply amending the definition of the macro `yy_set_interactive` using the m4 variables `M4_YY_PROTO_LAST_ARG` and `M4_YY_DECL_GUTS_VAR` already defined in `flex.skl`?

Are there any issues when doing so? Of course, I see the advantage of having a function instead of a macro, especially from another c file, however, macros are preferred by flex.

Separately, would it be fair to say that the remainder of the changes are enhancements to command line options only?

Explorer09:
@jannick0 The use of the function instead of a macro is because YY_CURRENT_BUFFER_LVALUE is an internal macro, AFAIK. That won't be the problem if yy_set_interactive is called only within the .l file, but that's not the case.

jannick0:
@Explorer09 Agree, but the issue then becomes a philosophical one, namely if the implementation should prefer macros (as is essentially right now) or functions. Using a wrapper would a way out of the current situation to effectively call macro defined in the lex file only without any changes.

The short patch above is amending what I would consider a bug, since for reentrant scanners the opaque scanner object needs to be resolved as it contains the flag `is_interactive` as a member to be set.

rmccullagh:
@jannick0 If you read #290, you can see that `yy_set_interactive` is perfectly callable from the `.l` file, but if you need more control, you don't have access to that macro because it is not defined in a header file. Furthermore, the required data structure to implement `yy_set_interactive` is an internal data structure.

src/flex.skl
rmccullagh: @Explorer09 Good call, I will get this updated tonight.

Explorer09:
There are weaknesses of C macro mainly because they work like simple string substitution. If you only use C macro for reasons of function inlining, please don't, or do it sparingly.

I should have mentioned that yy_set_interactive, in it's current macro form, lack the proper `do {...} while (0)` enclosure. See this (https://stackoverflow.com/questions/257418/do-while-0-what-is-it-good-for) for reasons.
Above is one of the reasons you should avoid macro as much as possible. Modern compilers are excellent at inlining functions, so they can inline `static` functions when that benefits. Speaking of static functions, I would like to bring issue #277 together with this, since I would like if there's a way to make yyset_interactive static when we have the function implemented.

src/flex.skl
rmccullagh: @Explorer09 This is done.

jannick0:
Again, using macros or functions is a question of coding style which is not my business here, supposedly rather the package maintainer's. Other issues about how to use macros are interesting, but to be dealt with in a different place I would say. Of course, I agree that they should certainly be followed for implemented macros.

If a function implementation of `yy_set_interactive` is ultimately chosen, then I would strongly recommend not defining it as `static` sucht that it can be made available in other entities. [BTW to avoid any confusion: Even the macro implementation of any procedure like  `yy_set_interactive` (not defined in the flex header file) can be made available in other entities using a public function wrapper defined in the .l file as long as the opaque scanner object type is available in that very entity. The latter is needed either way, the macro of the function implementation.]

PR #295, opened 2017-12-14
introduce yycalloc and apply to internal scanner structs

jannick0:
Use zero-initialized memory block for default initialization of struct members. If accepted

- the default initialization of struct members by 0 could be dropped.
- struct members added to the code later would be zero-initialized.

### DIFF
--- a/doc/flex.texi
+++ b/doc/flex.texi
@@ -2993,6 +2993,7 @@ list of the names affected:
     yytext
     yywrap
     yyalloc
+    yycalloc
     yyrealloc
     yyfree
 @end verbatim
@@ -4848,10 +4849,11 @@ you call yylex_init(). It is destroyed when the user calls yylex_destroy().
 @section Overriding The Default Memory Management
 
 @cindex yyalloc, overriding
+@cindex yycalloc, overriding
 @cindex yyrealloc, overriding
 @cindex yyfree, overriding
 
-Flex calls the functions @code{yyalloc}, @code{yyrealloc}, and @code{yyfree}
+Flex calls the functions @code{yyalloc}, @code{yycalloc}, @code{yyrealloc}, and @code{yyfree}
 when it needs to allocate or free memory. By default, these functions are
 wrappers around the standard C functions, @code{malloc}, @code{realloc}, and
 @code{free}, respectively. You can override the default implementations by telling
@@ -4867,6 +4869,7 @@ following options:
 @itemize
 @opindex noyyalloc
 @item @code{%option noyyalloc}
+@item @code{%option noyycalloc}
 @item @code{%option noyyrealloc}
 @item @code{%option noyyfree}.
 @end itemize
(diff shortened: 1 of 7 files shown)


Explorer09:
I wonder if this would bring performance problem for allocated data that doesn't need zero initialization, but filled in in other ways.
Look at `yy_scan_bytes` in the skeleton for an example, its `buf` is not zero-initialized, but the data is copied from `yybytes` argument.

westes:
We're already using calloc() in some places to ensure initialization occurs with predictable behavior. I'm not convinced that performance in these cases is worth worrying about.

Explorer09:
This patch is supposed to remove some memset() calls after yyalloc() in the skeleton, so it's already addressing performance in the first place.
I don't think using calloc() over malloc() will give a behavior any more predictable, though. This thinking won't be sufficient if you're going to address data safety.

jannick0:
`malloc` could then be replaced by `calloc` when it comes to memory allocation of internal flex structures. This is a bit more work you are suggesting, so this needs the implementation.

Explorer09:
@jannick0 
How about renaming all the so-called `yyalloc` to `yymalloc` or `yycalloc` to address separate needs?
I found another problem with the name `yyalloc`: Bison have been using this already. ([You can Google the word yyalloc and see its results](https://google.com/search?q=yyalloc).) It's defined as a union type.
We would better rename it as well if we are to change its behavior.

jannick0:
- Renaming `yyalloc` to `yymalloc` and `yycalloc` would imply some additional work in `flex.skl` and `scan.l` and the documentation. Of course, `yyalloc` should be labelled as deprecated before it could be fully removed to avoid any confusion of external code calling `yyalloc` for whatever reason. 

- Compilers do not have any issues with structures / unions and functions having the same name in one single compilation entity, so this is not a real problem. If a human code reader of bison's and flex' output married ot each other became confused because of that, flex' `%option prefix=...` could help him. There are tons of examples of bison and flex pairs out there which do not cause any issue of which flex is just one fundamental one. Given flex' naming logic and features, I would stick to `yymalloc`, `yycalloc` and - for backward compatibility supposedly for over some time - a deprecated version of `yyalloc`.

westes:
On Tuesday, 26 December 2017,  9:19 am +0000, jannick0 <notifications@github.com> wrote:

> - Renaming `yyalloc` to `yymalloc` and `yycalloc` would imply some additional work in `flex.skl` and `scan.l` and the documentation. Of course, `yyalloc` should be labelled as deprecated before it could be fully removed to avoid any confusion of external code calling `yyalloc` for whatever reason. 

Yes.

> - [...] Given flex' naming logic and features, I would stick to `yymalloc`, `yycalloc` and - for backward compatibility supposedly for over some time - a deprecated version of `yyalloc`.

Yes.

We need some sense of how widely yyalloc is used outside of flex itself. The first step in the process is to add the functions yymalloc and yycalloc and document their existence.

> -- 
> https://github.com/westes/flex/pull/295#issuecomment-353944072

-- 
Will Estes
westes575@gmail.com

jannick0:
@westes @Explorer09 Updated commits:
- yycalloc introduced
- yycalloc applied to internal scanner structs only, but not to memory allocation for buffers (unneccessary and due to performance reasons).

Hope that looks better then.

PR #297, opened 2017-12-17
change token name 'CHAR' to 'TOK_CHAR'

jannick0:
change avoids conflict with type definition of CHAR

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -1,6 +1,6 @@
 /* parse.y - parser for flex input */
 
-%token CHAR NUMBER SECTEND SCDECL XSCDECL NAME PREVCCL EOF_OP
+%token TOK_CHAR NUMBER SECTEND SCDECL XSCDECL NAME PREVCCL EOF_OP
 %token TOK_OPTION TOK_OUTFILE TOK_PREFIX TOK_YYCLASS TOK_HEADER_FILE TOK_EXTRA_TYPE
 %token TOK_TABLES_FILE
 
@@ -757,7 +757,7 @@ singleton	:  singleton '*'
 		|  '(' re ')'
 			{ $$ = $2; }
 
-		|  CHAR
+		|  TOK_CHAR
 			{
 			++rulelen;
 
@@ -788,7 +788,7 @@ braceccl:
 			}
 		;
 
-ccl		:  ccl CHAR '-' CHAR
+ccl		:  ccl TOK_CHAR '-' TOK_CHAR
 			{
 
 			if (sf_case_ins())
@@ -848,7 +848,7 @@ ccl		:  ccl CHAR '-' CHAR
 			$$ = $1;
 			}
 
-		|  ccl CHAR
+		|  ccl TOK_CHAR
 			{
 			ccladd( $1, $2 );
 			cclsorted = cclsorted && ($2 > lastchar);
@@ -927,7 +927,7 @@ ccl_expr:
 				}
 		;
 		
-string		:  string CHAR
+string		:  string TOK_CHAR
 			{
 			if ( $2 == nlch )
 				rule_has_nl[num_rules] = true;
--- a/src/scan.l
+++ b/src/scan.l
@@ -74,7 +74,7 @@ extern const char *escaped_qstart, *escaped_qend;
 
 #define RETURNCHAR \
 	yylval = (unsigned char) yytext[0]; \
-	return CHAR;
+	return TOK_CHAR;
 
 #define RETURNNAME \
 	if(yyleng < MAXLINE) \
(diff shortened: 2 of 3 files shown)


Explorer09:
What's anything wrong with the type definition of CHAR?

Explorer09:
What's anything wrong with the type definition of CHAR? It's a Flex internal symbol, and not available in generated scanners. Where's the necessity of renaming?

jannick0:
For now there is no conflict with the definition of `CHAR` in Windows' system header `winnt.h`. This change does not harm the flex code at all.

Explorer09:
@jannick0 
I would bring inconsistency in Flex tokens' naming. If it doesn't break, don't fix it.
I personally would suggest prefixing all Flex tokens with FLEX_TOK_ if we are to address name conflict problems in the future.

jannick0:
The `TOK_` prefix is already used in flex' bison grammar, so no additional inconsistency. Separately, I would agree to the longer version of a fully consistent prefix, however I would be concerned to make the bison grammar less readable then once the longer token names are used therein (given that that we cannot - or rather must not - expect a minimum version of bison for compiling flex.)

westes:
Yes, if we're going to indicate that something is a flex-centric thing, then we should use FLEX_ as the prefix.

On Tuesday, 26 December 2017,  9:28 am +0000, jannick0 <notifications@github.com> wrote:

> The `TOK_` prefix is already used in flex' bison grammar, so no additional inconsistency. Separately, I would agree to the longer version of a fully consistent prefix, however I would be concerned to make the bison grammar less readable then once the longer token names are used therein (given that that we cannot - or rather must not - expect a minimum version of bison for compiling flex.)
> 
> -- 
> https://github.com/westes/flex/pull/297#issuecomment-353945195

-- 
Will Estes
westes575@gmail.com

jannick0:
@westes Will, would you be happy to assume `bison >= 3.0` - with a predefined bison exit status otherwise? That would help using the bison feature to consistently use a token name prefix (here: `FLEX_TOK_`) without cluttering the bison grammar itself?

westes:
Yes.

On Tuesday, 26 December 2017,  8:57 am -0800, jannick0 <notifications@github.com> wrote:

> @westes Will, would you be happy to assume `bison >= 3.0` - with a predefined bison exit status otherwise? That would help using the bison feature to consistently use a token name prefix (here: `FLEX_TOK_`) without cluttering the bison grammar itself?
> 
> -- 
> https://github.com/westes/flex/pull/297#issuecomment-353989446

-- 
Will Estes
westes575@gmail.com

Explorer09:
I would prefer not to do it, since tracing all uses of the symbol would become difficult, and `find | xargs grep` would not find all the use of it.

Can we go back to the start and think of the purpose of this change? We use the prefix mainly to avoid name collision with other parts of the code, and potential collisions with user code, if they incorporate or embed Flex code to make a scanner _generator_ for their use.

The use of a unified prefix would also allow Flex symbols to stand unique in the Google search results, although this is only a side benefit.

I know you would concern about `FLEX_TOK_` being too long for a token prefix. But in this case there are alternatives. `_FLT_` could also work if we could claim the prefix for our exclusive use.

I think this issue can be deferred to the next Flex major release; existing code hasn't broke; so there is no need to rush. I can wait for more suggestions to come in before the maintainer makes a decision.

westes:
I'm fine with FLEX_TOK_ for the flex token stuff.

I would think you'd want to use ctags or similar to find all the occurences of a token, not find.

On Tuesday, 26 December 2017,  6:00 pm +0000, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> I would prefer not to do it, since tracing all uses of the symbol would become difficult, and `find | xargs grep` would not find all the use of it.
> 
> Can we go back to the start and think of the purpose of this change? We use the prefix mainly to avoid name collision with other parts of the code, and potential collisions with user code, if they incorporate or embed Flex code to make a scanner _generator_ for their use.
> 
> The use of a unified prefix would also allow Flex symbols to stand unique in the Google search results, although this is only a side benefit.
> 
> I know you would concern about `FLEX_TOK_` being too long for a token prefix. But in this case there are alternatives. `_FLT_` could also work if we could claim the prefix for our exclusive use.
> 
> I think this issue can be deferred to the next Flex major release; existing code hasn't broke; so there is no need to rush. I can wait for more suggestions to come in before the maintainer makes a decision.
> 
> -- 
> https://github.com/westes/flex/pull/297#issuecomment-353995460

-- 
Will Estes
westes575@gmail.com

jannick0:
Replaced by #299.

PR #300, opened 2017-12-28
continuous integration

westes:
There are a number of non-usual things about flex -- from the fact that most tool chains assume that flex exists, which isn't true when building flex itself -- to the fact that some parts of flex are just down right self referential. To aid contributors in assessing the effects of their changes, to aid the maintainer in assessing changes, to ensure that flex continues to do what we all think it does, let's start testing flex's build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: c
+
+compiler:
+  - gcc
+
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - autoconf
+    - bison
+    - gcc-6
+    - help2man
+    - lzip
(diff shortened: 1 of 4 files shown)


westes:
Note well that this is a work in progress. I am intentionally not doing it on master so that other work, if needed, will be uneffected. The expectation is that builds will fail until I sort out what all travis needs.

westes:
As expected, building flex with travis defaults fails. We don't want to require flex to bootstrap flex, however, so simply installing whatever the ubuntu default of "flex" is, while straight forward, is not correct.

The right solution is to build the most recent version of flex from source, install that as "flex" and then use that to bootstrap the tip of whatever branch we're using.

PR #301, opened 2017-12-29
misc.c: no YYTABLES_NAME macro unless requested

jannick0:
Like in all other flex output files the macro `#define YYTABLES_NAME "yytables"` is currently appearing in `stage1scan.c` regardless of the user's requests. The reason for the spurious macro is that there is no m4 guard for external tables in the skeleton, however this fix is in line with the approach how other external tables related statements are handled.

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -776,8 +776,9 @@ void skelout (void)
 					outn ((char *) (yydmap_buf.elts));
 			}
             else if (cmd_match (CMD_DEFINE_YYTABLES)) {
-                out_str("#define YYTABLES_NAME \"%s\"\n",
-                        tablesname?tablesname:"yytables");
+                if ( tablesext )
+                    out_str( "#define YYTABLES_NAME \"%s\"\n",
+                           tablesname ? tablesname : "yytables" );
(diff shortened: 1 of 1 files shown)


src/misc.c
```diff
@@ -776,8 +776,9 @@ void skelout (void)
 					outn ((char *) (yydmap_buf.elts));
 			}
             else if (cmd_match (CMD_DEFINE_YYTABLES)) {
-                out_str("#define YYTABLES_NAME \"%s\"\n",
-                        tablesname?tablesname:"yytables");
+		if(tablesext)
+                   out_str("#define YYTABLES_NAME \"%s\"\n",
+                           tablesname?tablesname:"yytables");
```
Explorer09: Sorry to bother. But can you fix the coding style a little bit as you fix the code here?
There should be a space between `if` and left parenthesis (in order to distinguish the syntax of function or macro call).  
It's also better to put spaces before and after each of the `? :` operators since they have low precedence in the language.

jannick0: Done.

westes:

PR #308, opened 2018-01-31
flexint: include stdint.h before defining limit macros

### LINKED ISSUE #307: duplicate definition of INT8_MAX in Visual Studio 2015
In generated source files, [this block](https://github.com/westes/flex/blob/master/src/flexint.h#L32) is done after including `inttypes.h`, however at least on Visual Studio 2015, these are defined by `stdint.h` which is included later in the file causing redefinition warnings.

mathstuf:
These defines are provided by `stdint.h` and it should be included first
before trying to define them manually.

Fixes #307.

---
Cc: @jannick0 @Explorer09

### DIFF
--- a/src/flexint.h
+++ b/src/flexint.h
@@ -30,6 +30,7 @@ typedef unsigned short int flex_uint16_t;
 typedef unsigned int flex_uint32_t;
 
 /* Limits of integral types. */
+#include <stdint.h>
 #ifndef INT8_MIN
 #define INT8_MIN               (-128)
 #endif


Explorer09:
No no. Please don't patch like this. There's no need to include `stdint.h`at all if `inttypes.h` is included. `inttypes.h` is more widely available than `stdint.h`, as far as we know, so Flex chose to include the former.
From what I found on the web, Visual Studio is one of the exceptions that did not provide `inttypes.h`, but has `stdint.h` since VS2010. I think you should do a MSC_VER check in your patch, so that we do this include workaround only for MSVC.

westes:
We want to check for features, not versions.

On Wednesday, 31 January 2018,  3:52 pm +0000, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> No no. Please don't patch like this. There's no need to include `stdint.h`at all if `inttypes.h` is included. `inttypes.h` is more widely available than `stdint.h`, as far as we know, so Flex chose to include the former.
> From what I found on the web, Visual Studio is one of the exceptions that did not provide `inttypes.h`, but has `stdint.h` since VS2010. I think you should do a MSC_VER check in your patch, so that we do this include workaround only for MSVC.
> 
> -- 
> https://github.com/westes/flex/pull/308#issuecomment-361975425

-- 
Will Estes
westes575@gmail.com

mathstuf:
`inttypes.h` is only included on the other side of the `#if __STDC_VERSION__` check; it isn't included at all on the patched codepath.

Explorer09:
#309 is my version of the patches.

mathstuf:
Closing in preference to #309.

PR #309, opened 2018-01-31
Remove integer limit macros from skeleton

### LINKED ISSUE #307: duplicate definition of INT8_MAX in Visual Studio 2015

Explorer09:
Resolves #307.
See descriptions of both commits.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -58,6 +58,7 @@ COMMON_SOURCES = \
 	filter.c \
 	flexdef.h \
 	flexint.h \
+	flexint_shared.h \
 	gen.c \
 	main.c \
 	misc.c \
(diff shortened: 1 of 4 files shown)


mathstuf:
Thanks, this looks better indeed. One thing I might ask is that `Visual C++ is not fully C99 compliant` instead be `Older Visual C++ compilers are…`, since VS2015's toolchain is C99-compliant now (since C++11 requires it).

Explorer09:
@mathstuf Okay. I'll correct the commits if I have time.

westes:

Explorer09:
#329 is a regression of this pull.

GitMensch:
> Thanks; this is on master and will be included in the next release.

Sigh, just stumbled over this ... again. So still waiting for "the next release" to land. 🥺 🙏

Issue #313, opened 2018-02-14
lex.yy.c: error: implicit declaration of function 'strdup'


AdrianBunk:
Commit a17d79e9 breaks code that requires functionality from POSIX standards more recent than 1990, in this case `strdup`:

https://buildd.debian.org/status/fetch.php?pkg=libreswan&arch=arm64&ver=3.23-2&stamp=1518540444&raw=0

```
parser.l: In function 'parser_y_init':
parser.l:104:33: error: implicit declaration of function 'strdup' [-Werror=implicit-function-declaration]
  ic_private.stack[0].filename = strdup(name);
                                 ^~~~~~

```
With gcc in non-strict mode (which is the default, and the libreswan build above even explicitely sets -std=gnu99) _POSIX_C_SOURCE defaults to 200809.
Commit a17d79e9 defines _POSIX_C_SOURCE to 1, and this commit was already backported into flex in Debian unstable.
`strdup` requires _POSIX_C_SOURCE >= 200809.

I do not think flex should mess with feature test macros in the files it generates.

Explorer09:
You should #define _POSIX_C_SOURCE to an appropriate value in your user code. I suggest you fix it in the `libreswan` package downstream.

You can add something like
```
%top{
#define _POSIX_C_SOURCE 200809L
}
```

I think this is due to our lack of documentation, I apologize if this is the case. (Can you suggest a documentation fix?)

AdrianBunk:
Yes, `libreswan` and other packages that got broken due to your change could workaround this regression by starting to also mess around with feature test macros. This doesn't mean that it is correct that `flex` suddenly started to overwrite the compiler setting in generated files.

Note that the packages that fail to build are only the tip of the iceberg: Without the `-Werror` your change would have resulted in `libreswan` being miscompiled, and there are likely packages that are getting miscompiled due to that for obvious reasons.

`flex` has been using `fileno` without messing with feature test macros for > 20 years, and this is also documented.

Explorer09:
@AdrianBunk 
I think I found the problem... it's my fault, definitely, when I miss the glibc <features.h>.

Explorer09:
Would you please test the patch in #315?
BTW, I think `libreswan` and other packages (and flex also) should definitely consider adding the flag `-Werror=implicit-function-declaration` even on release builds.

AdrianBunk:
#315 looks clearly bogus.

Quoting `feature_test_macros(7)`:
[...]

PR #315, opened 2018-02-19
scanner: Fix glibc features.h dependency in skeleton

Explorer09:
Commit a17d79e9c722a6735b6d2a8f152287404f27df32 defines `_POSIX_C_SOURCE`
to the minimum of 1 if it's not defined in the user's scanner code or
the compiling environment. However in glibc the macros are not yet set
up until one of the libc headers is included. This unfortunately have
made us overwrite the default `_POSIX_C_SOURCE` value that would be
defined by glibc (`200809L` at the time of writing), causing regressions
in user code.

Now in this patch:
1. Ensure feature test macros have been set up in glibc before checking
   or defining any of them in our skeleton code.
2. Have a more conservative logic when determining the need to define
   `_POSIX_C_SOURCE` (required for fileno()).

Fixes: #313

Note:
It could be tricky for application code to ensure feature test macros
have been set up in glibc, since <features.h> is no portable header and
not meant to be included directly by applications. The way to do it is
to include a libc header which in turn includes <features.h>. However,
many of the glibc headers check `__USE_POSIX` (a glibc internal macro
defined if `_POSIX_C_SOURCE` is defined) and determine which interfaces
to expose already, making the headers inappropriate for our goal.
Those which don't depend on `_POSIX_C_SOURCE`, and are also available
since ANSI C89, are only <assert.h>, <errno.h> and <math.h>.

<assert.h> is finally favored due to other considerations:
- <math.h> check for `__USE_XOPEN` in glibc, making a dependency on
  `_XOPEN_SOURCE`, besides it exposes much more interfaces than we need.
- In djgpp, <errno.h> depends on `_POSIX_SOURCE` to hide definitions of
  some errno values when it's defined.
- <assert.h> exposes the fewest interfaces among the 3 headers and, at
  the time of writing, checks for only C99 (for `__func__`), C11 (for
  `_Static_assert`), and `_GNU_SOURCE` when needed. (And the header had
  been available in as old as V7 Unix.)

### DIFF
--- a/src/flex.skl
+++ b/src/flex.skl
@@ -220,11 +220,21 @@ m4_ifdef( [[M4_YY_TABLES_EXTERNAL]],
 %if-c-only
 m4_ifdef( [[M4_YY_ALWAYS_INTERACTIVE]], ,
 [[m4_ifdef( [[M4_YY_NEVER_INTERACTIVE]], ,
-[[#ifndef _POSIX_C_SOURCE
-#define _POSIX_C_SOURCE 1 /* for fileno() */
-#ifndef _POSIX_SOURCE
-#define _POSIX_SOURCE 1
+[[/* Feature test macros. Flex uses functions that require a minimum set of
+ * macros defined. As defining some macros may hide function declarations that
+ * user code might use, be conservative and respect user's definitions as much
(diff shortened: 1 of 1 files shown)


westes:

PR #316, opened 2018-02-22
build: Move dnl comments out of AC_CHECK_FUNCS

Explorer09:
Due to a bug, autoheader (2.69) will treat M4 dnl comments in a quoted
argument of AC_CHECK_FUNCS as function tokens and generate a lot of
redundant and useless HAVE_* macros in config.h.in.
(Examples: HAVE_DNL, HAVE_AVAILABLE_, HAVE_BY)

It seems to be this commit dbb4e94dc7bacbcfd4acef4f085ef752fe1aa03f of
mine that revealed this autoheader bug, and the affected config.h.in
had been shipped within flex-2.6.4 release tarball.

I have reported the autoheader bug here:
<https://lists.gnu.org/archive/html/bug-autoconf/2018-02/msg00005.html>

As a workaround, let's move comments out of AC_CHECK_FUNCS.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -158,22 +158,19 @@ AC_FUNC_REALLOC
 AS_IF([test "$cross_compiling" = yes],
    AC_MSG_WARN([result $ac_cv_func_realloc_0_nonnull guessed because of cross compilation]))
 
-AC_CHECK_FUNCS([dup2 dnl
-memset dnl
-regcomp dnl
-strcasecmp dnl
-strchr dnl
-strdup dnl
-strtol dnl
-], [], [AC_MSG_ERROR(required library function not found on your system)])
+dnl Autoheader (<= 2.69) bug: "dnl" comments in a quoted argument of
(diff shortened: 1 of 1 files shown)


westes:

egorpugin:
@westes 
Hi! Any chances moving to cmake?

(not proposing my big c++ related changes, but build system only - as I remember cmake config was in good shape last year)

westes:
My recollection is that cmake is missing things like gettext support and the kind of function checking that we're doing.

If I'm wrong, please point me at the relevant docs, although I'd suggest doing that on a separate issue especially since this one is now closed.

On Thursday,  1 March 2018,  5:01 am -0800, Egor Pugin <notifications@github.com> wrote:

> @westes 
> Hi! Any chances moving to cmake?
> 
> (not proposing my big c++ related changes, but build system only - as I remember cmake config was in good shape last year)
> 
> -- 
> https://github.com/westes/flex/pull/316#issuecomment-369584407

-- 
Will Estes
westes575@gmail.com

Issue #311, opened 2018-02-06
Travis CI failures on gpg keyserver error


Explorer09:
Travis CI build system (as used in flex) can fail when a GPG key is not found in a keyserver.
However it should not fail because the fault is not with the developers / patch submitters, and it would be confusing when a red X is suddenly shown in GitHub's pull request, (For a fault that truly isn't theirs.)

westes:
So the solution is to include the gpg keys in the travis build setup. Could you submit a pull request to do that for the existing scripts?

Explorer09:
For future reference, these two are examples of build server failures simply caused by GPG server fail:
https://travis-ci.org/westes/flex/jobs/335672015
https://travis-ci.org/westes/flex/jobs/338037463

PR #317, opened 2018-02-23
travis: Hard-code hashes for verifying tarballs

Explorer09:
GPG signatures require an external keyserver which might be offline,
which is undesirable for build server use. It's equally secure to just
hard-code the hashes, provided they're trusted (i.e. you verify a hash
against a GPG signature once).

Fixes: #311.

As a side note: the original two signatures
(gettext-0.19.8.1.tar.lz.sig and automake-1.15.1.tar.gz.sig) are signed
against the files' SHA-1 hash.

### DIFF
--- a/.travis/install-automake.sh
+++ b/.travis/install-automake.sh
@@ -1,8 +1,18 @@
 #!/bin/bash -ex
 
-wget -nv https://ftp.gnu.org/gnu/automake/automake-1.15.1.tar.gz{,.sig}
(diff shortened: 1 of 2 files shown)


Explorer09:
Also note that I have also made the verification optional (the `|| :` syntax makes it continue even after a hash mismatch).

westes:
Since this is superceded by #318, I am closing.

PR #318, opened 2018-02-23
travis: Silence build output of automake and gettext

Explorer09:
There is a lot of output when building gettext and automake, the two dependent packages need to build Flex in Travis CI. Their output are not useful in diagnosing Flex problems, so just silence them unless a build failure comes from one of them.


westes:

PR #320, opened 2018-02-25
build: $(WARNINGFLAGS) appended to AM_CFLAGS as a rule

Explorer09:
In src/Makefile.am, remove flex_CFLAGS variable, and append
$(WARNINGFLAGS) to AM_CFLAGS. The use of flex_CFLAGS has forced
Automake to rename flex's object files (.o) to have a 'flex-' prefix.
By preventing the rename, the generated src/Makefile.in can be smaller
(-35kB, 30%).

For now this commit brings only the benefit of a smaller Makefile, I
plan to simplify the process of the native build of flex by sharing
generated .o files whenever possible between 2 stages (so only the
scanner and linked executable will need to be rebuilt in stage 2). And
this commit will be a prerequisite for that goal.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,5 +1,6 @@
 AM_YFLAGS = -d
 AM_CPPFLAGS = -DLOCALEDIR=\"$(localedir)\"
+AM_CFLAGS = $(WARNINGFLAGS)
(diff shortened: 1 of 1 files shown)


Explorer09:
Closed in favor of #321.

PR #321, opened 2018-02-28
New cross-compiling and bootstrap mechanism.

### LINKED ISSUE #314: Not cross-compiling correctly
Some more fun with cross-compiling...

$ make stage1flex
gcc  -DHAVE_CONFIG_H -I. -I../../flex-2.6.4/src  -DLOCALEDIR=\"/usr/share/locale\" -isystem/data/poky-tmp/master/work/corei7-64-poky-linux/flex/2.6.4-r0/recipe-sysroot-native/usr/include  -isystem/data/poky-tmp/master/work/corei7-64-poky-linux/flex/2.6.4-r0/recipe-sysroot-native/usr/include -O2 -pipe -c -o stage1flex-misc.o `test -f 'misc.c' || echo '../../flex-2.6.4/src/'`misc.c
../../flex-2.6.4/src/misc.c: In function ‘allocate_array’:
../../flex-2.6.4/src/misc.c:151:14: warning: implicit declaration of function ‘reallocarray’ [-Wimplicit-function-declaration]
  new_array = reallocarray(NULL, (size_t) size, element_size);
              ^~~~~~~~~~~~
...
stage1flex-misc.o: In function `allocate_array':
misc.c:(.text+0x36f): undefined reference to `reallocarray'

It took me a minute to figure this out but the problem is that my *target* has reallocarray support but my *host* does not.  config.h says the function exists so misc.c tries using the reallocarray() (implicit declaration) and then undefined reference at link time.

My suggestion is to stop trying to cross-build a stage1flex and just expect anyone building a cross-compiled flex to bring their own recent flex binary for the build host.

Explorer09:
TODO:
- [x] Rebase (done, not pushed yet)
- [ ] Need a workaround for the $(BUILD_OBJEXT) bug, because $(BUILD_OBJEXT) in AX_PROG_CC_FOR_BUILD doesn't actually work for now. (And need to report it upstream, too.)

----
1. Deterministic stage1scan.c.
2. Introduce config_for_build.h for cross, bootstrap config. Fixes #314.
3. Native builds of flex are now faster because most object codes (except the scanner code) are shared between stage 1 and stage 2 builds.
4. For cross builds, BUILD_EXEEXT and BUILD_OBJEXT should now work on first stage. That is, crossing from Linux to Windows should be possible (although untested).
5. No longer requires GNU Make to start a cross + bootstrap build. (GNU Make syntax avoided.)


westes:
@Explorer09 you labeled this as experimental. Do you think it's ready to go into what will be 2.6.5 or would you like more time, say "2.6.6"?

Explorer09:
@westes I personally think it better to be in 2.6.6. Although I tested all of the commits by myself, I just feel there have been too many changes here for 2.6.5.

westes:
There, it's assigned to the 2.6.6 milestone.

Explorer09:
P.S. I can show an example of what I would fear of. The sedding mechanism on makefile (4df74badca7bd1b9f7b3c1cfc599d1ce0c8ccfa9) is a new trick I made, but during testing, I accidentally found out BusyBox sed (1.21.1, Ubuntu 14.04) could fail with it due to [a bug](https://git.busybox.net/busybox/commit/?id=b0e9b72915a87121aabf21d4e382e16e0a9a5175), and I barely made the detection and workaround code. Who would know where this new cross-building way would break? I wish to have more time to find out.

Explorer09:
Rebased and refreshed for a few bugfixes:

* The malloc(0) cross compiling check did not respect user settings of the ac_cv_func_malloc_0_nonnull variable.
* The scan.c generation in dist-hook is now also deterministic. (Previously it would present `#line 1 "../src/scan.l"` if a builder builds the tarball from a VPATH build directory.)
* sed regressions caused by M4 quoting rules (missing necessary quotes) and configure script peculiarities.

Explorer09:
I think I need to close this pull request as I revise the commits. This branch has been filled with commits unrelated to the main goal of making of "new bootstrap mechanism", and I think it's better split them to separate PRs (which could be reviewed and merged faster than the main one), so that this branch can focus on one thing.

PR #323, opened 2018-03-04
scanner: Fix scan.l xgettext warnings.

Explorer09:
xgettext was not very clever at interpreting lex patterns and can get
confused when seeing unquoted quotation marks, and emit warnings for
them. Now fix the warnings by properly quoting the quotation marks in
lex regex patterns.

Example
    Original: `\"[^"\n]*\"` -> "warning: unterminated string literal"
    Fixed:    `"\""[^""\n]*"\""` -> OK

My basic build test shows that the generated stage1scan.c is
bit-identical to the original.


westes:

PR #324, opened 2018-03-12
Makefile minor fixes

Explorer09:
1. Overwriting the files via `mv` command needs write permission on the directory, but the write permission on the file is not necessary. (In contrast, overwriting via `cp` command needs write permission on the file; write permission on the directory is optional.)
2. Make `mv` command more portable by specifying `-f` flag.


westes:

PR #327, opened 2018-03-16
Update README.md

GitMensch:
Changes concerning the mailing lists...

I actually wrote a longer note to find out some hours later that the message was not send because I missed to subscribe - this would be nice to know before sending to the mailing list...


westes:

Issue #322, opened 2018-03-01
CMake support


egorpugin:
Referring https://github.com/westes/flex/pull/316#issuecomment-369586876

gettext was implemented here:
https://github.com/egorpugin/flex/blob/master/po/CMakeLists.txt
(no cmake docs here, it's just a small script)

Different checks goes here:
https://github.com/egorpugin/flex/blob/master/cmake/Configure.cmake#L41
See my similar contribution here:
https://github.com/FirebirdSQL/firebird/blob/master/builds/cmake/Configure.cmake#L199

CMake docs for checking:
https://cmake.org/cmake/help/v3.10/module/CheckFunctionExists.html
https://cmake.org/cmake/help/v3.10/module/CheckIncludeFile.html
and (see Check*.cmake files)
https://cmake.org/cmake/help/v3.10/manual/cmake-modules.7.html

Explorer09:
IMHO, it is good to see flex to support building with CMake. However we are withing for volunteers to do it because at least I don't have knowledge with CMake syntax.

By the way, note that I have been recently tweaking the configure scripts and makefiles in order to improve cross-compile procedures (see #321). I just wish that CMake-files could also catch up. :)

westes:
Outside of the technical issues, we would need to verify that the various OS's -- particularly the BSD's -- include cmake in their basic/bootstrap systems.

On Thursday,  1 March 2018,  4:44 pm +0000, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> IMHO, it is good to see flex to support building with CMake. However we are withing for volunteers to do it because at least I don't have knowledge with CMake syntax.
> 
> By the way, note that I have been recently tweaking the configure scripts and makefiles in order to improve cross-compile procedures (see #321). I just wish that CMake-files could also catch up. :)
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/westes/flex/issues/322#issuecomment-369653694

-- 
Will Estes
westes575@gmail.com

egorpugin:
True.
But it's possible to co-exist both build systems for some time.
Early adopters could try to switch to cmake early.
Others will wait for their experience and switch later.
As I understand, bsd systems have cmake.

Explorer09:
@egorpugin 

> My port already exists for almost two years. We need only administrative decision.

[...]

PR #328, opened 2018-03-19
Update README.md, changed instructions to use a separate build directory

GitMensch:
Reasons for doing so:

* it is more clean
* I actually got a strange `configure: error: source directory already configured; run "make distclean" there first` error message when trying to configure from srcdir, running configure from a separate builddir dir work without any issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,9 @@ GNU-style configure script to be able to work.
 From this point on, building flex follows the usual routine:
 
 ```bash
-configure && make && make install
+mkdir build
+cd build
+../configure && make && make install
 ```
 
 ---


Explorer09:
@GitMensch 
Sorry, but this issue of yours didn't make sense. The `source directory already configured` shouldn't appear if you run configure in source directory (it's supposed to allow running multiple times; what it prohibits you from is configuring in source and out of source simultaneously).
You can see the comments in #322, that we don't like to require users to configure out of source directory.

GitMensch:
@Explorer09 
> Sorry, but this issue of yours didn't make sense. The `source directory already configured` shouldn't appear if you run configure in source directory (it's supposed to allow running multiple times; what it prohibits you from is configuring in source and out of source simultaneously).

I know that it doesn't make any sense - but this happens on the machine where I test the master branch. Note: this doesn't happen with all other tools build on this machine (which were quite a few...). If you have any ideas how to debug this: I'm open to cooperate ;-)

> You can see the comments in #322, that we don't like to require users to configure out of source directory.

It is good to not *require* users to do anything specific and I wouldn't like this special *requirement* as a user myself. But *if* users follow the instructions I'd consider it a bonus to show them a way that should always work (note: I know that it doesn't work for all packages which is bad [but it works with flex]).

Explorer09:
@GitMensch 
As this PR is closed, I'm sorry to ask: Are you going to file an issue report about your `source directory already configured` problem?
With reproduce steps or a brief list of the directory contents (the Unix `find` command could help) when the error appeared.

GitMensch:
@Explorer09 I'll try to do this next week, but this may also be partially rooted in the way how I got the package (download as zip vie GitHub, changed gettext version, chmod+w two files, used old environment to bootstrap, ....)
I suggest that you do a fresh checkout and bootstrap + configure + make dist and upload the tarball here as this would allow me to test with a reasonable "user" tarball and checking if the next release can be build on this system without an extra build directory... before the release. [I could check this on Tuesday]

PR #330, opened 2018-03-19
Update autogen.sh, add executable right for po/update_linguas.sh

GitMensch:
Not sure if there are more to include here. The missing rights may produced by wrong copying, using GitHub's download as zip function or by using GitHubs svn bridge (I assume they are set as executable with the available git function).

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -49,3 +49,8 @@ if ! test -f ChangeLog; then
 fi
 "$LIBTOOLIZE" --install --force
 autoreconf --install --force
+
+# add executable right for auxiliary scripts
+if test -f po/update_linguas.sh; then
+   chmod gu+x po/update_linguas.sh
+fi


Explorer09:
Why do we need executable permission for the script if the user has no need to run it? If they need to run it, why not just let them `chmod u+x` themselves?

GitMensch:
At least when build from VCS (which is the main reason to run `autogen.sh`) the user likely want to run it. And in this case it is likely that the missing execution bits are set.

As it doesn't "cost" anything with the benefit that the user *could* run it I'd say its a nice (even though not "needed") addition to the autogen script.

westes:
That file is already executable in the git repo. If the github zip links don't preserve that, that is likely intentional on github's part. That has never been a desired or supported method of acquiring flex sources from my point of view -- if i could turn it off I would.

Why would anyone use a git->svn bridge for flex? Flex went straight from cvs into git.

GitMensch:
> That file is already executable in the git repo.

That is good.

>If the github zip links don't preserve that, that is likely intentional on github's part. That has never been a desired or supported method of acquiring flex sources from my point of view -- if i could turn it off I would.

The same may happen with a bad file copy, too.

> Why would anyone use a git->svn bridge for flex? Flex went straight from cvs into git.

Actually this is likely to happen when someone want to get the sources which is unfamiliar with git but familiar with svn (or works on a machine which has a subversion but no git client installed, which may happen on a server of a corporate network more often than you think...).
As GitHub says when you click on "clone or download" button:

> Use Git or checkout with SVN using the web URL.

It is a simple `svn co https://github.com/westes/flex.git/trunk` to get master while an installation of git (especially on "old" systems) beforehand may be much more difficult.

But of course, you are free to close the PR without merge, it was mainly a suggestion to remove one manual step to get the updated translations.

... speaking of these: would you accept a PR to include the script in the release tarball?

westes:
On Tuesday, 20 March 2018, 12:36 am -0700, Simon Sobisch <notifications@github.com> wrote:

> The same may happen with a bad file copy, too.

Then the file copy should be corrected.

> Actually this is likely to happen when someone want to get the sources which is unfamiliar with git but familiar with svn (or works on a machine which has a subversion but no git client installed, which may happen on a server of a corporate network more often than you think...).

Then the correct step would be to become familiar with git since all flex development happens in a git-centric world. As to the corporate network, it is possible to clone over http/s, so network traffic is not the issue. If the corporate controls do not permit a git client to be run, then the corproate controls are the problem.

> As GitHub says when you click on "clone or download" button:
> 
> > Use Git or checkout with SVN using the web URL.
> 
> It is a simple `svn co https://github.com/westes/flex.git/trunk` to get master while an installation of git (especially on "old" systems) beforehand may be much more difficult.
> 
> But of course, you are free to close the PR without merge, it was mainly a suggestion to remove one manual step to get the updated translations.

I see this pr as making changes to accomodate broken things outside of flex which is in general a bad idea. autogen.sh should do the minimum necessary to get configure/make working and no more.

> ... speaking of these: would you accept a PR to include the script in the release tarball?

Yes, that is reasonable.

Explorer09:
Just a few notes...

westes
> If the github zip links don't preserve that, that is likely intentional on github's part. That has never been a desired or supported method of acquiring flex sources from my point of view -- if i could turn it off I would.

The original zip format does not store Unix file permissions (i.e. the file modes), and so it's quite normal for a script to become not executable after extracting from zip. The equivalent may also happen if you copy flex source from a noexec mount (i.e. a USB drive in FAT32 filesystem).

GitMensch
> Actually this is likely to happen when someone want to get the sources which is unfamiliar with git but familiar with svn (or works on a machine which has a subversion but no git client installed, which may happen on a server of a corporate network more often than you think...).

If I remember, `svn export` does set executable permissions for scripts. Even if it doesn't, that's not our problem to address anyway.

Explorer09:
P.S. I finally realize how absurd this patch is, when the script that sets executable permission for other scripts needs executable permission of itself to run. A user has to `chmod u+x autogen.sh` anyway if the script is not executable --- it's same amount of effort for user to  `chmod u+x update_linguas.sh`.

westes:
On Tuesday, 20 March 2018,  3:43 am -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> The original zip format does not store Unix file permissions (i.e. the file modes), and so it's quite normal for a script to become not executable after extracting from zip. The equivalent may also happen if you copy flex source from a noexec mount (i.e. a USB drive in FAT32 filesystem).

Good points. And yet, flex is a utility for the unix-like world, so compensating for bugs/misfeatures/dissimilarities makes flex have cruft for no benefit to flex itself. The solution is to get flex sources in a manner compatible with flex sources.

GitMensch:
> If I remember, `svn export` does set executable permissions for scripts (or else how do you run ./configure, obviously). Even if it doesn't, that's not our problem to address anyway.

Yes, `svn` handles this correctly, but not in this case as the subversion bride used by GitHub doesn't map any git attributes to svn attributes.

> A user has to `chmod u+x autogen.sh` anyway if the script is not executable --- it's same amount of effort for user to `chmod u+x update_linguas.sh`.

Yes, this is true and actually the reason that I've voted myself for closing this (and deleted the change-branch). It is what I've done, too (note: all other files are taken care for by the tools used by the autogen script).

PR #332, opened 2018-03-20
Update Makefile.am

GitMensch:
include po/update_linguas.sh to allow the user to update the translations before installing - as discussed in https://github.com/westes/flex/pull/330

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -39,7 +39,8 @@ dist_doc_DATA = \
 
 EXTRA_DIST = \
 	.indent.pro \
-	autogen.sh
+	autogen.sh \
+	po/update_linguas.sh
 
 SUBDIRS = \
 	src \


GitMensch:
@westes Hm. Added to the milestone for the next release but not merged? [this applies to more than one PR: https://github.com/westes/flex/milestone/4]

westes:
Issues and pull requests are marked for inclusion in a release generally before work on the issue/pr is done. It just means I haven't gotten to it yet.

westes:
Is there a reason you didn't create an additional Rules-* file in po/?

https://www.gnu.org/software/gettext/manual/gettext.html#po_002fRules_002d_002a

GitMensch:
> Is there a reason you didn't create an additional Rules-* file in po/?

I just haven't used them before. And even when I would I'd do it the same way as:

* providing a shell (I guess you didn't meant to include the update script in the Makefile, did you?) means the user can update even in non-configured environments
* I rather include a single line than creating another file only for including the script - if there's more to be found than this single line a Rules file would be appropriate

westes:
The problem with using the top level Makefile.am is that it's not where I'd expect to find a change to how po/ is handled. Normally you'd use the Makefile.am in the directory, but since it's gettext, the mechanism there is Rules-*.

On Saturday,  7 April 2018, 12:01 pm -0700, Simon Sobisch <notifications@github.com> wrote:

> > Is there a reason you didn't create an additional Rules-* file in po/?
> 
> I just haven't used them before. And even when I would I'd do it the same way as:
> 
> * providing a shell (I guess you didn't meant to include the update script in the Makefile, did you?) means the user can update even in non-configured environments
> * I rather include a single line than creating another file only for including the script - if there's more to be found than this single line a Rules file would be appropriate
> 
> -- 
> https://github.com/westes/flex/pull/332#issuecomment-379491606

-- 
Will Estes
westes575@gmail.com

westes:
So it turns out that automake is a lot more hands off about gettext than I thought. I ended up doing two things:

- listing update_linguas.sh as you had suggested in this pr.
- using dist_noinst_SCRIPTS to get the scripts to have executeable permissions.

Thanks for the pr. I'm closing it and the change as I have described it will be in the next release of flex, albeit under my id since it took a bit more poking at to figure out how best to do it.

GitMensch:
Yes, 4c7e3a5 is fine.

Issue #340, opened 2018-04-16
Memory leaks in xstrdup


xcainiao:
./flex ./flex_memory_leaks_in_xstrdup

git log
```
commit 60b578fdaab4bf8d95203caaa947b476eeb03c10
Author: Simon Sobisch <simonsobisch@web.de>
Date:   Fri Mar 16 21:53:48 2018 +0100

    doc: describe mailing list usage in more detail
```
compile
```
./configure CC="gcc" CXX="g++" CFLAGS="-g -fsanitize=address"
make
```
error
```
==110424==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 2 byte(s) in 1 object(s) allocated from:
    #0 0x7f320858230f in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x6230f)
    #1 0x420712 in xstrdup /home/fuzz/github/flex_o/src/misc.c:231
    #2 0x435dd2 in ndinstal /home/fuzz/github/flex_o/src/sym.c:180
    #3 0x441ee7 in flexscan scan.l:342
    #4 0x43d4c0 in yylex /home/fuzz/github/flex_o/src/yylex.c:51
    #5 0x427ca7 in yyparse /home/fuzz/github/flex_o/src/parse.c:1405
    #6 0x41edea in readin /home/fuzz/github/flex_o/src/main.c:1490
    #7 0x41af93 in flex_main /home/fuzz/github/flex_o/src/main.c:170
    #8 0x41b282 in main /home/fuzz/github/flex_o/src/main.c:209
    #9 0x7f3207e6d82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)

Direct leak of 1 byte(s) in 1 object(s) allocated from:
    #0 0x7f320858230f in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x6230f)
    #1 0x420712 in xstrdup /home/fuzz/github/flex_o/src/misc.c:231
    #2 0x435dc3 in ndinstal /home/fuzz/github/flex_o/src/sym.c:180
    #3 0x441ee7 in flexscan scan.l:342
    #4 0x43d4c0 in yylex /home/fuzz/github/flex_o/src/yylex.c:51
    #5 0x427ca7 in yyparse /home/fuzz/github/flex_o/src/parse.c:1405
    #6 0x41edea in readin /home/fuzz/github/flex_o/src/main.c:1490
[...]

PR #341, opened 2018-04-17
xstrdup memory leak

xcainiao:
https://github.com/westes/flex/issues/340


Explorer09:
@xcainiao 
No, please don't. The free() statements you made in 9800a5003e7f726257a42af7bd1ba23eebea106b would break the "ownership" assumptions in addsym(), and it's a bad idea.
Specifically, since addsym() did not perform strdup's on `sym` or `str_def` argument, it shouldn't be its responsibility to free it; the responsibility should belong to callers.
If we fail to follow this assumption, a dangerous **use after free** could occur.

westes:
Closing as per @Explorer09

PR #342, opened 2018-04-19
bug fix: bracket parsing in %top{...} code blocks 

jannick0:
The bracket parsing of %top{...} has flaws in flex 2.6.4 fixed by the package.

Unfortunately the test file top.l is (almost) completely replaced albeit some lines with sample code blocks for bracket parsing are added.

@westes Will, please amend as you find appropriate, the edits to the manual, in particular.

### DIFF
--- a/src/scan.l
+++ b/src/scan.l
@@ -137,7 +137,7 @@ extern const char *escaped_qstart, *escaped_qend;
 
 %x SECT2 SECT2PROLOG SECT3 CODEBLOCK PICKUPDEF SC CARETISBOL NUM QUOTE
 %x FIRSTCCL CCL ACTION RECOVER COMMENT ACTION_STRING PERCENT_BRACE_ACTION
-%x OPTION LINEDIR CODEBLOCK_MATCH_BRACE
+%x OPTION LINEDIR CODEBLOCK_TOP
 %x GROUP_WITH_PARAMS
 %x GROUP_MINUS_PARAMS
 %x EXTENDED_COMMENT
@@ -195,7 +195,7 @@ STRING_LITERAL   \"(\\\"|\\{NL}|[^"\r\n])*\"
                 ++linenum;
                 buf_linedir( &top_buf, infilename?infilename:"<stdin>", linenum);
                 brace_depth = 1;
-                yy_push_state(CODEBLOCK_MATCH_BRACE);
+                yy_push_state(CODEBLOCK_TOP);
             }
 
     ^"%top".*   synerr( _("malformed '%top' directive") );
@@ -297,7 +297,7 @@ STRING_LITERAL   \"(\\\"|\\{NL}|[^"\r\n])*\"
 			}
 }
 
-<CODEBLOCK_MATCH_BRACE>{
+<CODEBLOCK_TOP>{
     "'}'"      |
     "'{'"      |
     [^{}\r\n]  |
@@ -305,6 +305,8 @@ STRING_LITERAL   \"(\\\"|\\{NL}|[^"\r\n])*\"
 
     {STRING_LITERAL}  |
     {C_COMMENT}     LINENUM_UPDATE; buf_strnappend(&top_buf, yytext, yyleng);
+
+    ^"%}"      buf_strnappend(&top_buf, "\n", 1); yy_pop_state();
 
     "}"     {
                 if( --brace_depth == 0){
--- a/src/scan.l
+++ b/src/scan.l
@@ -123,6 +123,13 @@ extern const char *escaped_qstart, *escaped_qend;
     if (!indented_code) line_directive_out(NULL, 0);\
 } while (0)
 
+#define LINENUM_UPDATE do {         \
+    int yy1 = 0;                    \
+    while(yy1 < yyleng)             \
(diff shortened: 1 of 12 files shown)


Explorer09:
Which bug does this fix? Can we have a test code that shows an incorrect behavior and what behavior is expected?

jannick0:
Everything already there: run `top.l` after 7cdd24380b144873c6a352f8f27e66e70e37fd1d against flex 2.6.4. See commentaries in `top.l`.

src/scan.l
```diff
@@ -285,9 +285,24 @@ M4QEND      "]""]"
 }
 
 <CODEBLOCK_MATCH_BRACE>{
+    "'}'"      |
```
Explorer09: Can you simplify `"'{'" | "'}'"` to this?
```
'[{}]'
```

src/scan.l
```diff
@@ -285,9 +285,24 @@ M4QEND      "]""]"
...
+    "'{'"      |
+    [^{}\r\n]  |
+    \"(\\\"|[^"\r\n])*\"  |
```
Explorer09: I think in C, backslash in end-of-line to make multi-line string literal is allowed.
```
"something like\
this"
```

src/scan.l
```diff
@@ -285,9 +285,24 @@ M4QEND      "]""]"
...
+    "//".*     buf_strnappend(&top_buf, yytext, yyleng);
+
+    [/][*]+([^*]|([*][^/]))*[*]+[/] {
```
Explorer09: I wonder if this syntax will match `/*` marks quoted in a literal.
```
"example /* " "/*" } "*/"
```
I think you get the idea.

jannick0:
@Explorer09:
ad 1 - Agree. May I kindly ask you to file PRs as you say you prefer incl. the suitably adjusted testsuite such that we can review?

ad 2 - Well, the question is rather if flex is able to treat a `%top{ ... }` code block like a `%{ ... %}` for which the answer is no. The main reason behind the enhancement is the weird syntax that the closing tag of a top code block is a simple closing bracket instead of `%}` (at BOL). For consistency reasons I would have expected `%top{ ... %}` instead.

src/scan.l
jannick0: Yes - what does this mean in terms of table size and performance?

src/scan.l
jannick0: Correct, but fairly seldomly used. flex will report an error to user in this case who could then simply adjust the code. We should not overcomplicate things here and target a perfect parser given that no user filed a complaint about that. I am opting to not tackle all edgy corners of the c/c++ syntax.

But we are happy to review any enhancement.

src/scan.l
jannick0: Yes - a seemingly tricky one, but assuming it is stand-alone it should by broken into three string literals with `}` before the last one (other white spaces ignored).  I have not tested. Did you?

src/scan.l
Explorer09: I would argue that all of your cases you made in this PR can all be worked-around. That why I ask you for real world code that has been broken by the "bug" you noted in the title.

jannick0: It is clear that this is a bug, since according to the manual top code blocks are same as `%{ ... %}` code blocks. For a real life example put code documentation into the top section which somewhere contains a `}`.

Not sure which workarounds you are thinking of, but if valid code cannot be put into the top code section (even without any far-reaching syntax exceptions) causing the user to fiddle around is still a bug.

westes:
Whatever the eventual soution is, yes, flex 3.0 would be the place to introduce an incompatibility like that.

On Friday, 20 April 2018, 10:22 am -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> @jannick0 
> I would suggest this: Let's deprecate `%top{ ... }` code block unless the closing brace is obviously in the same line as the beginning one. For a multi-line `%top` block, emit a warning for using `}` instead of `%}` for closing. That way we would make Flex syntax more defined.
> 
> However there's a backward compatibility problem and I don't think it's good to rush this one until a major release (maybe flex 3.0.0). 
> 
> -- 
> https://github.com/westes/flex/pull/342#issuecomment-383165084

-- 
Will Estes
westes575@gmail.com

jannick0:
@Explorer09 What you are suggesting would probably cause to break existing code out there because of a syntax change. I am not in favor of the one-line solution either if things should remain simple. As suggested above, the cleanest version is `%top{ ... %}` with opening and closing tag sitting at BOL. Could you then provide a PR for your suggestion such that we can review if needed? It would be best if both syntax versions can coexist for a while, where the current one could potentially go if decided so at all. Thanks.

@westes Of course, I want to leave the decision with you, Will.

westes:
The one line idea would run counter to reasonable expectations of users of flex.

Using %} at BOL as the closing for a %top block is probably what we want for parsing reasons. That is incompatible and should be delayed till either flex 2.7.0 or 3.0.0. Having both syntaxes available in the meantime is acceptable since it is backwards compatible.

jannick0:
@westes : 
- Unindented %} as closing tag of %top implemented in fbd46ec.
- this PR should go nicely with PR #352

Issue #268, opened 2017-09-29
flex `%option noline` outputs a #line directive


fooofei:
## Used flex version:
2.6.4 
download from https://github.com/westes/flex/releases/download/v2.6.4/flex-2.6.4.tar.gz

## Input file
I have a file which filename is `test.flex`,
the content is 
```
%option noline
%option c++

%%
.* {return 1;}
%%
```

## Process

I used command line `flex test.flex` to generate file `lex.yy.option_noline.cc`,

then I remove the `%option noline` from file `test.flex`,

I use command line `flex --noline test.flex` to generate file `lex.yy.commandline_noline.cc`.

git diff the two generate file, a part of the diff is:
```
diff --git a/lex.yy.commandline_noline.cc b/lex.yy.option_noline.cc
index 893d434..2678b1e 100755
--- a/lex.yy.commandline_noline.cc
+++ b/lex.yy.option_noline.cc
@@ -386,9 +386,15 @@ static const flex_int16_t yy_chk[8] =
 #define yymore() yymore_used_but_not_detected
 #define YY_MORE_ADJ 0
 #define YY_RESTORE_YY_MORE_OFFSET
+#line 1 "test.flex"
+
+
[check_lineno.zip](https://github.com/westes/flex/files/1342843/check_lineno.zip)

 ```

We can see, if we use `%option noline`, flex also outputs `#line` to the generate `.cc` file ( `lex.yy.option_noline.cc` ) which was not expected. On the other hand, the command line option `flex --noline` is not outputs `#line`.

## Found & Research:
I found in the flex source code, it use a global variable `int gen_line_dirs` to decide flex is outputs
`#line` or not. but when flex outputs `#line` the variable  `gen_line_dirs` is 1, then it make outputs.

I debug the source code, the callstack is:
(gdb) bt
#0  add_action (new_text=0x7fffffffd170 "#line 1 \"test.flex\"\n") at misc.c:134
#1  0x000000000040ed93 in line_directive_out (output_file=0x0, do_infile=1) at misc.c:379
#2  0x000000000040d798 in readin () at main.c:1473
#3  0x000000000040ae18 in flex_main (argc=2, argv=0x7fffffffe2f8) at main.c:170
#4  0x000000000040af6d in main (argc=2, argv=0x7fffffffe2f8) at main.c:209

The line number at the end of callstack line maybe not right, I add some `fflush(stdout)` in the file for debug, it will affect the line number.

releated issues of this repo:
https://github.com/we

[...]

PR #344, opened 2018-04-19
fix logic to (not) create #line directive

jannick0:
addressing #268 (again).


Smattr:
@jannick0, could I ask why you closed this? AFAICT this resolves #268 which is still an issue. Thank you for posting this fix and I was looking forward to it being merged.

Issue #296, opened 2017-12-15
flex combined configuration (from cmd line and %options) not invariant under permutation


jannick0:
After parsing all command line arguments and `%option`s a configuration consistency check is not performed.

**Examples** - While
```c
%option yywrap main
%%
.
```
nicely compiles,
```c
%option main yywrap
%%
.
```
does not due to a missing reference to `yywrap`, since the boolean variable `do_yywrap` should be superseded by the option of `main`. This happens in the first case (scan.l), but not in the second. 

Could possibly be implemented in `flex.skl` using `m4` or in flex code (after both the command line flags and the %options are parsed). Probably variable `do_main` to be introduced.

There might be other consistency checks needed to avoid a malformed configuration for which an error could be thrown.

jannick0:
Suggestion to add something like this
```c
if ( do_main ) do_yywrap = false;
```
to the consistency checks in the function `check_options()` in `main.c`. While scanning a flex input file it is called when hitting the first `%%`. The global boolean variable `do_main` would then be required to be added with default `false`.

Issue #346, opened 2018-04-20
command line flag --main ignored


jannick0:
Running a minimal example against flex 2.6.4 with command line flag `--main` does *not* produce a `main` function albeit requested. However the `%option main` does work.

jannick0:
closed on the back of PR #347

PR #347, opened 2018-04-20
consistent treatment of '--main' and '%option main'

jannick0:
fixing bug #346 where command line flag `--main` is ignored when running `flex --main mini.l` with
```c
 /* mini.l */
%%
.
```
flex 2.6.4 does not provide a default main function in the generated code, although it does so if `%option main` is used instead.

### DIFF
--- a/src/flex.skl
+++ b/src/flex.skl
@@ -3412,9 +3412,9 @@ int main ()
 
 %if-reentrant
     yyscan_t lexer;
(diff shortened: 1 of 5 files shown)


jannick0:
This PR fixes #296, since the option consistency check `main` vs. `yywrap` is now moved into function `check_options`. 

@westes  #296 could be closed if/when this PR is merged into master. Thx!

PR #348, opened 2018-04-21
test suite: updates and edits to pacify compiler

jannick0:


westes:
From skimming the commit messages, this is more properly several smaller pull requests.

Please split this up accordingly.

In general, one pr should fix one test.

The obvious exception to that is if you're doing something to make things consistent as in one of your commit messages. That's ok to submit as one pr spanning multiple tests. But there should be nothing else in that pr.

PR #353, opened 2018-04-24
build: Fix unportable sed expression in mkskel.sh

### LINKED ISSUE #294: mkskel.sh & Apple sed
Since 3f2b9a4 it only works with GNU sed.

On Apple sed (BSD too?) it breaks lines at `r`, producing invalid C file.

Explorer09:
'\r' in sed is not portable, at least to BSD sed (and macOS).
Replace it with embedded CR character in sed expression.

Fixes #294 (regression caused by #243)

@stek29 and @jannick0, I need you to test this one as I don't have a BSD sed to do so.


jannick0:
@Explorer09  That looks good for LF and CRLF files where `sed` from cygwin32 and msys2 is called. I cannot say anything about CR files, since I am on Windows, but I am curious if it works on macOS, as well.

Explorer09:
@jannick0 I'm not caring for CR-only files yet. Since Mac OS Classic ports of `sed` (if any) would need to treat CR as _pattern space_ terminators anyway and is beyond what our sed expression can workaround already.
My idea was to accept both CR+LF and LF, but output in the OS-native line termination scheme (or whatever the `sed` was configured to output). Any complex line termination scheme (CR only, NEL (`'\x85'`) or form feeds (`'\f'`)) is beyond the scope of this patch.

Explorer09:
Note: `sed` uses the _pattern space_ concept instead of a line for processing. The reason is that `sed` allows joining multiple lines into one _pattern space_ for processing together (with `N` command), or split a line into two spaces (`P;D` commands combination) for processing. The _pattern space_ is more flexible than a line in terms of stream processing.

stek29:
Sorry for long wait
I confirm it works fine with Apple (BSD?) sed.

westes:
Thanks for the confirmation.

On Friday, 27 April 2018, 12:09 pm +0000, Viktor Oreshkin <notifications@github.com> wrote:

> Sorry for long wait
> I confirm it works fine with Apple (BSD?) sed.
> 
> -- 
> https://github.com/westes/flex/pull/353#issuecomment-384951452

-- 
Will Estes
westes575@gmail.com

wendajiang:
Resolve #539

PR #354, opened 2018-04-24
Fix {ccl,nd,sc}instal functions memory leak & other imprevements

Explorer09:
Move the job of strdup() from cclinstal(), ndinstal() and scinstal() to
addsym(). This means ccltab[], sctbl[] and ndtbl[], respectively, will
assume ownership of the duplicated string buffers, and no more garbage
will be left in cclinstal(), ndinstal() or scinstal().

Because of the ownership assumption of sctbl[], strings in scname[] are
now READ-ONLY as they point to the "name" members within sctbl[].

Fixes: #340


Explorer09:
I have forgotten how or why I made this pull request and this one has been years ago. I don't know if the original issue still exists. Maybe someone else can review the code again for me.

PR #356, opened 2018-05-03
build: add __attribute__s for deprecating functions etc.

jannick0:
As promised here a package providing features for deprecating and obsoleting functions (plus more). It effectively relies on `__attribute__`s for functions (and variables).

1. add m4 macro templates (from the autoconf-archive) for checking function and variable `__attribute__`s and have `autotools` process them
2. add macro definitions `FLEX_ATTRIBUTE_FUNCTION_...` and `FLEX_ATTRIBUTE_VAR_...` to `flexdef.h`
3. sample application to label functions
4. example of how to deprecate a function (here: `yy_flex_xmalloc`). Suggested messages and flex version number can certainly be adjusted.

With this it is easy to add new attributes to the system.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,6 +56,24 @@ AC_PROG_LN_S
 AC_PROG_AWK
 AC_PROG_INSTALL
 
+#
+# check for func and var __attribute__(()) variations
+#
+AX_GCC_FUNC_ATTRIBUTE([alloc_size])
+AX_GCC_FUNC_ATTRIBUTE([always_inline])
+AX_GCC_FUNC_ATTRIBUTE([deprecated])
+AX_GCC_FUNC_ATTRIBUTE([format])
+AX_GCC_FUNC_ATTRIBUTE([format_arg])
+AX_GCC_FUNC_ATTRIBUTE([gnu_inline])
+AX_GCC_FUNC_ATTRIBUTE([malloc])
+AX_GCC_FUNC_ATTRIBUTE([nonnull])
+AX_GCC_FUNC_ATTRIBUTE([noreturn])
+AX_GCC_FUNC_ATTRIBUTE([unused])
+
+AX_GCC_VAR_ATTRIBUTE([deprecated])
+AX_GCC_VAR_ATTRIBUTE([unused])
+
+
 pkgconfigdir=${libdir}/pkgconfig
 AC_SUBST(pkgconfigdir)
 
--- a/configure.ac
+++ b/configure.ac
@@ -22,8 +22,15 @@
 # WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
 # PURPOSE.
 
+#
+# flex version read from '.prev-version' and format checked
+#
+m4_define( [FLEX_VERSION], [m4_esyscmd_s([cat .prev-version])])
+m4_define( [RX_FLEX_VERSION], [^\([0-9]+\)\.\([0-9]+\)\.\([0-9]+\)$])
+m4_bmatch( FLEX_VERSION, m4_defn([RX_FLEX_VERSION]), [],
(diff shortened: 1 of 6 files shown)


jannick0:
Solely checking for `__attribute__` is not sufficient. However, an additional existence check for each attribute is needed, since they were gradually introduced with the compiler development. Does your approach ensure that *each* of the requested attributes exists for the used compiler - regardless of its version?

Happy to see easier solutions, of course.

jannick0:
Your assumption is that for **each** attribute `A` a version pair `(__GNUC__, __GNUC_MINOR__)` is known such that for **each** compiler out there `A` is implemented. I don't have this list on a reliable basis, do you? This list would then go into the code which is not my aim.

In order to be on the super-safe side here and to really avoid breaking the compilation of existing code, I decided not to go down this route as you are suggesting and rather implement checks on the `configure` level which is sufficiently general to work for any compiler out of the box.

What is wrong with that given that `flex` is widely used for all sorts of compilers? Regarding the issue of multi-compiler compilation I am very much inclined not to use in-src-compilation, but build-dir, for which a unique `configure` for all compilers would perfectly fit. I am convinced that in-src is not efficient. (NB: I have not looked into your PR in depth, but had a short glance.)

westes:
In general, the autotools approach is to check for feature presence/absence/functionality, rather than to check for specific programs and versions.

On Thursday,  3 May 2018,  3:31 pm -0700, jannick0 <notifications@github.com> wrote:

> Your assumption is that for **each** attribute `A` a version pair `(__GNUC__, __GNUC_MINOR__)` is known such that for **each** compiler out there `A` is implemented. I don't have this list on a reliable basis, do you? This list would then go into the code which is not my aim.
> 
> In order to be on the super-safe side here and to really avoid breaking the compilation of existing code, I decided not to go down this route as you are suggesting and rather implement checks on the `configure` level which is sufficiently general to work for any compiler out of the box.
> 
> What is wrong with that given that `flex` is widely used for all sorts of compilers? Regarding the issue of multi-compiler compilation I am very much inclined not to use in-src-compilation, but build-dir, for which a unique `configure` for all compilers would perfectly fit. I am convinced that in-src is not efficient. (NB: I have not looked into your PR in depth, but had a short glance.)
> 
> -- 
> https://github.com/westes/flex/pull/356#issuecomment-386456685

-- 
Will Estes
westes575@gmail.com

Explorer09:
On Thursday, 3 May 2018, 3:31 pm -0700, jannick0
<notifications@github.com> wrote:
> Your assumption is that for each attribute A a version pair (__GNUC__, __GNUC_MINOR__) is known such that for each compiler out there A is implemented. I don't have this list on a reliable basis, do you? This list would then go into the code which is not my aim.

Yes. Even if we don't know it at the moment, we can look it up on the internet
later. And Clang have helped us by introducing `__has_attribute()` so we can
check for features also for non-GNU compilers.

How are these insufficient?

Also, all of the `__attribute__`s are GNU extensions to C, not part of standard
C. They are not required to compile flex for most functionality, although they
can provide useful optimizations and warning checks.

> In order to be on the super-safe side here and to really avoid breaking the compilation of existing code, I decided not to go down this route as you are suggesting and rather implement checks on the configure level which is sufficiently general to work for any compiler out of the box.

Not if some developers take out the flex code and incorporate it in software
that does not use Autotools.

jannick0:
Additional commits on the way to a deprecation and obsoletion process:
* config: add `FLEX_VERSION_MAJOR/MINOR/SUBMINOR` to `config.h`
* config: retrieve flex version from .prev-version in the format major.minor.subminor
* flexdef.h: new macro `FLEX_CHECK_VERSION` to compare version (x,y,z) to flex version
* example for deprecation and obsoletion of a function

Issue #67, opened 2016-03-22
flex: yy_fatal_error could print location in reentrant lexers


srivasta:
Hi,

  This feature request was submitted by a Debian user in:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=603787

since/when the running lexer (not only reentrant) does record the location,  please make yy_fatal_error print it by default.

  Thank you,

  Manoj

PR #357, opened 2018-05-09
add line number to message in yy_fatal_error

jannick0:
As requested in #67. 
The fatal error message is enhanced by the line number derived from yylineno, but only if the scanner is generated with the option `yylineno` (for both reentrant and non-reentrant scanners).
CAVEAT: The line number info might be misleading if `yylineno` is incorrect.

### DIFF
--- a/src/flex.skl
+++ b/src/flex.skl
@@ -2478,7 +2478,13 @@ static void yynoreturn yy_fatal_error YYFARGS1(const char*, msg)
 {
 	M4_YY_DECL_GUTS_VAR();
 	M4_YY_NOOP_GUTS_VAR();
+m4_ifdef( [[M4_YY_USE_LINENO]],
+[[
+	fprintf( stderr, "fatal error in line %d: %s\n", yylineno, msg );
+]],
+[[
 	fprintf( stderr, "%s\n", msg );
+]])
 	exit( YY_EXIT_FAILURE );
 }
 %endif


Explorer09:
Please remove `M4_YY_CALL_LAST_ARG` from the fprintf calls. If you read its macro definition, you know that is meant for passing scanner states in a reentrant scanner, and it's not used in the way you did.

jannick0:
@Explorer09 Thanks! ... The funny part about that is that travis did not complain.

Explorer09:
I'm worrying about the format of the error message here. Although we did output line numbers in the error messages in this PR, the line number without the accompanying input file name is meaningless. Was there any way to retrieve  the input file name as well?

Explorer09:
@jannick0 
> I disagree that the line number alone is not meaningful. The user will be happy to know which code line causes the fatal error, despite the fact that the file name is not explicit.

Line numbers might be freely overridden with directives! For generated files, line directives may be used to indicate the lines copied from the source so that compilers or parsers (of whatever language) can indicate source file name and line numbers for diagnostic purposes.

You should expect that in scanners generated by Flex, users _may_ override `yylineno` for whatever purpose, including the use case above. Hence `yylineno` alone is not meaningful enough in diagnostic.

>It would helpful to see your suggestion for an error statement with/without file name?

See how gcc emits error messages, of course! Or C assertions, both of which at least emit both a file name and a line number for diagnostic!

westes:
Yes, the error message should emit file and line number since there could easily be more than one scanner in a given project.

If a user chooses to override the meaning of yylineno, it's on them to make it something useful to them.

The only edge case is if the file is standard input, but then you just print something meaningful for that.

On Saturday, 19 May 2018,  9:21 am -0700, jannick0 <notifications@github.com> wrote:

> * I disagree that the line number alone is not meaningful. The user will be happy to know which code line causes the fatal error, despite the fact that the file name is not explicit.
> * The file name is not available in the generic case of generated flex code. It is up to the user of flex to provide the file name if applicable.
> * It would helpful to see your suggestion for an error statement with/without file name?
> 
> -- 
> https://github.com/westes/flex/pull/357#issuecomment-390415717

-- 
Will Estes
westes575@gmail.com

Explorer09:
@westes I'm a bit confused now. Regarding line numbers, I think it ought to be two line number values during scanning:

1. The line number of _(potentially combined) scanner input_. It's internal and determines which line to fetch next from the file if the input was not from stdin. Overriding it would disrupt the scanning process of the flex-generated scanner.
2. The line number used for diagnostic output, this would be accompanied with the input file name, and both the name and line number may be overridden by user code just like how C and flex itself handle their `#line` directives from their input.

Which meaning was intended for yylineno in this regard?

westes:
yylineno is the 2nd of those.

On Saturday, 19 May 2018,  6:42 pm -0700, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> @westes I'm a bit confused now. Regarding line numbers, I think it ought to be two line number values during scanning:
> 
> 1. The line number of _(potentially combined) scanner input_. It's internal and determines which line to fetch next from the file if the input was not from stdin. Overriding it would disrupt the scanning process of the flex-generated scanner.
> 2. The line number used for diagnostic output, this would be accompanied with the input file name, and both the name and line number may be overridden by user code just like how C and flex itself handle their `#line` directives from their input.
> 
> Which meaning was intended for yylineno in this regard?
> 
> -- 
> https://github.com/westes/flex/pull/357#issuecomment-390451519

-- 
Will Estes
westes575@gmail.com

jannick0:
@westes Here a suggestion to what you said extended by the case of a string fed into a scanner.


* The feature is only available under `%option yylineno` where the  line number refers to `yylineno` which is under the user's control.
*  The text emitted to stderr would be  `%s:%d: fatal error\n, 'filename as defined below', yylineno`

1.   `yyin == NULL`: string and filename = `"<string>"`
2.  `yyin == stdin`: filename = `"<stdin>"`
3.   else: filename is to equal to macro `FLEX_THIS_FILE_ESC` (name derived from `__FILE__`) defaulting to the quoted and escaped path of the flex output file; `FLEX_THIS_FILE_ESC` is `#undef`ed at the end of the scanner file. This to be implemented in the skeleton plus infrastructure.

Agree? Or any objections, extensions or amendments to this? Happy to extend to current PR for this.

westes:
I think it's going to have to be in this pr, as much as I like to keep pr's small.

It looks reasonable, but others should have a chance to weigh in.

On Sunday, 20 May 2018,  2:29 am -0700, jannick0 <notifications@github.com> wrote:

> @westes Here a suggestion to what you said extended by the case of a string fed into a scanner.
> 
> 
> * The feature is only available under `%option yylineno` where the  line number refers to `yylineno` which is under the user's control.
> *  The text emitted to stderr would be  `%s:%d: fatal error\n, 'filename as defined below', yylineno`
> 
> 1.   `yyin == NULL`: string and filename = `"<string>"`
> 2.  `yyin == stdin`: filename = `"<stdin>"`
> 3.   else: filename is to equal to macro `FLEX_THIS_FILE_ESC` (name derived from `__FILE__`) defaulting to the quoted and escaped path of the flex output file; `FLEX_THIS_FILE_ESC` is `#undef`ed at the end of the scanner file. This to be implemented in the skeleton plus infrastructure.
> 
> Agree? Or any objections, extensions or amendments to this? Happy to extend to current PR for this. 
> 
> -- 
> https://github.com/westes/flex/pull/357#issuecomment-390468733

-- 
Will Estes
westes575@gmail.com

PR #358, opened 2018-05-10
flex.skl: add closing default statement in switch environment

### LINKED ISSUE #103: still some warnings with -Wswitch-default
Hello (again),

The following piece of code (see below) is ok under
flex (2.6.1 or 2.6.2) but still produces annoying warnings
when given to the -Wswitch-default option of gcc (i'm using gcc-6.2.0):

cm2.cc: In function 'int yylex()':
cm2.cc:814:15: warning: switch missing default case [-Wswitch-default]

cm2.cc: In function 'int yyinput()':
cm2.cc:1140:11: warning: switch missing default case [-Wswitch-default]

Addition of a simple 'default: break;' at various places inside the
flex code (2.6.1 or 2.6.2) corrects the problem.

%{
%}
%S INSIDE
%%
"beg of inside" { ECHO; BEGIN INSIDE; };
<INSIDE>"end of inside" { ECHO; BEGIN INITIAL; };
%%

jannick0:
fixes #103.

### DIFF
--- a/src/flex.skl
+++ b/src/flex.skl
@@ -1416,6 +1416,9 @@ do_action:	/* This label is used only to access EOF actions. */
 				yy_cp = YY_G(yy_c_buf_p);
 				yy_bp = YY_G(yytext_ptr) + YY_MORE_ADJ;
 				goto yy_find_action;
+
+			default:
+				break;
 			}
 		break;
 		}


Explorer09:
Such a little and seemingly obvious change makes me wonder: Was th6e `default` statement ever be reached? If it was unreachable code, an `unreachable` macro would be better.
Give us more explanation about the commit, and don't change the code just to workaround compiler warnings.

jannick0:
The code remains syntactically and semantically unchanged, so no need for you to wonder. What would be helpful for us, however, is to see your suggested macro for unreachable code if so.

Explorer09:
@jannick0 

> The code remains syntactically and semantically unchanged, so no need for you to wonder.

Um, excuse me, but what was the semantic? Does this empty default statement mean "do no-op if none matched in the list of cases", or "this default was in fact never reached, so compiler should optimize out the last JMP instruction". Which did you mean?

PR #364, opened 2018-06-18
scanner: Re-add 'bool' fallback definition.

Explorer09:
(Split from PR #321)

The fallback definition of 'bool' was deleted in
3efb40df0ec5cddf0cb63a7900eb88258c478514, and it was a bad idea to
require C99 stdbool.h for builders.


Explorer09:
Closed in favor of #373 (which includes this change)

PR #372, opened 2018-07-11
build: Deterministic scan.c and bootstrap byte comparison check

Explorer09:
(Split from PR #321)

Change the generating rules of scan.c (in dist-hook) and stage1scan.c
so that the source and output file names in #line directives are fixed.
This would allow the scanner results be comparable to each other.

Also in --disable-bootstrap build, the flex binary will now be built
with scan.c directly - no more sedding.

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -10,8 +10,10 @@ libfl.pc
 parse.c
 parse.h
 scan.c
-stage1scan.[cl]
+stage1scan.c
 stage1flex
+stage2compare
+stage2scan.c
 
 # for MSWindows
 
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -51,7 +51,11 @@ flex_SOURCES = \
 nodist_flex_SOURCES = \
 	$(SKELINCLUDES)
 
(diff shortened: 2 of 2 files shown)


Explorer09:
Update (2020-02-15): With stage1scan.c being deterministic, I added a byte comparison check (stage2scan.c) here that ensures the generated flex program can build the scanner of itself flawlessly.

Explorer09:
Updated according to @jannick0 's comments. The new version of this pull request now preserves stage2scan.c even if comparison fails. I don't think generating a diff output is good as it could be lengthy. The file preserved so if you want to make a diff you could do so yourself.

Explorer09:
Hello. It has been years after I made this pull request. Is Will (or other Flex maintainers) still interested in this? I may try to rebase it to the latest main branch if needed.

westes:
I'm working my way through the backlog (combination of easy and most important first).

I am 100% interested in this, yes.

westes:
Could you update this to account for #640?

Explorer09:
Rebased just now.

PR #385, opened 2018-09-16
added missing EXEEXT in tests

GitMensch:
allowing systems like Win32 (directly or cross-compilation) to actually compile and run the tests

tested with old sh (from MinGW) and GNU bash, Version 4.3.11(1)-release on Trisquel


GitMensch:
I'll just add another PR for gitignore and leave the other parts as they were...

PR #387, opened 2018-09-16
remove hard-wired bison lookup, check for "GNU Bison" in $YACC's version output instead

### LINKED ISSUE #386: hard bison name dependency
The current configure.ac has:

AC_PROG_YACC

which leads to `configure --help` saying:

>   YACC        The `Yet Another Compiler Compiler' implementation to use. Defaults to the first program found out of: `bison -y', `byacc', `yacc'.

So far everything is fine, but then configure.ac borks it by:

```
AS_IF([test "$YACC" != 'bison -y'], [
	YACC="\${top_srcdir}/build-aux/missing bison -y"
	AC_MSG_NOTICE(no bison program found: only required for maintainers)
	])
AM_CONDITIONAL([HAVE_BISON], [test "$YACC" = 'bison -y'])
```

Which results in this branch as soon as somebody does something like
`../configure YACC=/mydir/bison -y`.

GitMensch:
… fixes #386 

additional:
* don't check for GNU indent if indent-check already found no indent
* use common wording to check "*whether* $BINARY is GNU stuff"

You may pass a full path and/or a different name for the Bison executable and/or additional options for the bison executable now.

Tested with Cygwin and Trisquel.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -42,11 +42,18 @@ AC_SUBST(SHARED_VERSION_INFO)
 AM_GNU_GETTEXT([external])
 AM_GNU_GETTEXT_VERSION([0.19.7])
 AC_PROG_YACC
-AS_IF([test "$YACC" != 'bison -y'], [
-	YACC="\${top_srcdir}/build-aux/missing bison -y"
-	AC_MSG_NOTICE(no bison program found: only required for maintainers)
-	])
-AM_CONDITIONAL([HAVE_BISON], [test "$YACC" = 'bison -y'])
+  # Extract the first word of "$YACC", so it can be a program name with args.
+  set dummy $YACC; ac_word=$2
+  AC_MSG_CHECKING(whether $ac_word is GNU Bison)
+  AS_IF([test -n "$YACC" && $YACC --version 2>&1 | $GREP "GNU Bison" >/dev/null],
+    [AC_MSG_RESULT(yes)
+  	use_gnu_bison=1],
+    [AC_MSG_RESULT(no)
+  	AC_MSG_WARN($YACC does not appear to be GNU Bison; required for maintainers)
+  	use_gnu_bison=0
+  	YACC="\${top_srcdir}/build-aux/missing bison"
+    ])
+AM_CONDITIONAL([HAVE_BISON], [test $use_gnu_bison -eq 1])
 AM_PROG_LEX
 AC_PROG_CC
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -42,11 +42,18 @@ AC_SUBST(SHARED_VERSION_INFO)
...
 AM_GNU_GETTEXT_VERSION([0.19.7])
 AC_PROG_YACC
-AS_IF([test "$YACC" != 'bison -y'], [
-	YACC="\${top_srcdir}/build-aux/missing bison -y"
-	AC_MSG_NOTICE(no bison program found: only required for maintainers)
-	])
-AM_CONDITIONAL([HAVE_BISON], [test "$YACC" = 'bison -y'])
+  # Extract the first word of "$YACC", so it can be a program name with args.
+  set dummy $YACC; ac_word=$2
+  AC_MSG_CHECKING(whether $ac_word is GNU Bison)
```
Explorer09: I would like you to not bother with `$ac_word` if all you want to do is invoke `$YACC --version`. This would make code a bit smaller. The message `checking whether bison -y is GNU Bison` would be still sufficient for indicating what configure is doing.

GitMensch: The message `checking whether bison -y is GNU Bison` would be still sufficient for indicating what configure is doing.
Yes, it would, just looks a bit strange and is different from normal configure outputs. If requested I'll revert that part.

@westes your call: Do you want me to revert that part making configure.ac smaller (it doesn't count for the generated configure script which is much bigger) or stay with what the generated configure script normally does?

westes:
On Monday, 17 September 2018,  1:04 am -0700, Simon Sobisch <notifications@github.com> wrote:

> @westes your call: Do you want me to revert that part making configure.ac smaller (it doesn't count for the generated configure script which is much bigger) or stay with what the generated configure script normally does?

Please do so, yes.

GitMensch:
lines removed

GitMensch:
I know this is only 40 months old, but as we have a positive review by both @westes and @Explorer09 and a thorough discussion in the related issue - can you please merge this?

GitMensch:
Good that other issues were solved by merging PRs that had either a positive review or an actual request to be made (or both, like this one).... in the pure hope that this may be merged I'm sending out a ping @westes again.

GitMensch:
Friendly ping @westes as I'm cleaning up my branches...

PR #389, opened 2018-09-16
fix bad test results (false-positives) with versions that create windows-eol

GitMensch:
I see no reason that anything could break by this change (it just removes "\r" from an expected number) and it leads to more passing tests when using cl.exe and win_flex.exe.

### DIFF
--- a/tests/testwrapper.sh
+++ b/tests/testwrapper.sh
@@ -36,8 +36,8 @@ TESTNAME=$1
 INPUT_NAME=${INPUT_NAME:-$INPUT_DIRECTORY/`basename "${TESTNAME%.exe}"`.txt}
 
 if [ $DO_COMPARISON = 1 ] ; then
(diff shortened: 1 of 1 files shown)


Explorer09:
Why bother removing "\r"? EOL mishandling should be raised as en error, shouldn't it?
How was this thing causing failure to your test and why was this worth fixing?

GitMensch:
> Why bother removing "\r"? EOL mishandling should be raised as en error, shouldn't it?

No. You get those when running Windows based compiler+flex through POSIX-based shells like you have with Cygwin and possibly also MinGW.

> How was this thing causing failure to your test and why was this worth fixing?

It leads to the script evaluation of `"number\r" -eq "number\r"` - as `\r` is no number the `-eq` is invalid (another possible fix is to directly compar the values but I *guess* the test wanted to compare the numbers because of possible leading zero).

It is worth fixing to don't break "for no reason" when testing with these environments, especially as there is no side-effect (other than `tr` needed).

PR #408, opened 2018-11-30
Tighten up C source code scanning

DemiMarie:
* Parse C99/C++ line comments correctly
* Error if trigraphs present and could affect compilation.
* Reject invalid escaped newlines in many places
* Reject escaped newlines in a comment start or end, or in a line
  comment.


DemiMarie:
The main motivation for this change is that many (if not most) tools (that are not compilers) mishandle these constructs, and this probably includes other implementations of Lex.

westes:
There's been a lot of activity which has produced conflicts. Could you rebase against master and resolve the conflicts?

PR #410, opened 2018-11-30
Less m4

DemiMarie:
This changes some uses of M4 to use Flex’s skeleten file parsing machinery.

### DIFF
--- a/src/scan.l
+++ b/src/scan.l
@@ -49,48 +49,54 @@ extern const char *escaped_qstart, *escaped_qend;
 #define ESCAPED_QSTART "[" M4QEND M4QSTART "[" M4QEND M4QSTART
 #define ESCAPED_QEND M4QEND "]" M4QSTART M4QEND "]" M4QSTART
 
-#define ACTION_ECHO add_action( yytext )
-#define ACTION_ECHO_QSTART add_action (ESCAPED_QSTART)
-#define ACTION_ECHO_QEND   add_action (ESCAPED_QEND)
+#define ACTION_ECHO do add_action(yytext); while(0)
+#define ACTION_IFDEF(def, should_define) \
+	do if (should_define) action_define(def, 1); while (0)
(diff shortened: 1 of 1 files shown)


DemiMarie:
I suspect that `skelout()` will *always* return `true`, but I am not certain.

Explorer09:
Can you remove the regenerated po files from the first commit in this PR? That would make us reviewing the code easier.

DemiMarie:
@westes would you mind reviewing this?

westes:

DemiMarie:
> Can you see about the conflicts that have ben introduced from the merging in of #489?

Will do.  Given that we seem to be moving towards having more code in m4 (not less), I recommend that we embed a micro-m4 implementation ourselves, and have proper APIs for it.

DemiMarie:
Update: The machinery that I was trying to move Flex to has all been removed, so this PR now only contains the changes to C source code scanning.  C++11 raw string literals and Go raw strings still need to be handled, but that will come later.

westes:
The ultimate goal is no m4. What we need to be able to do that is to have uniformity in the code, so for now, we've got more m4 as that will be easier to replace, albeit down the road.

Mightyjo:
Oh! Good to know. I'd been happy with m4 as an intermediate language and was heading in an m4-heavy direction. This is a good time to course correct.

How do you imagine the endgame? Would m4 be gone entirely, relegated to the skeletons, abstracted behind backend classes, other?

westes:
Absolutely no m4 at all, anywhere.

Mind you, I have no thoughts on how we get there from here at the moment.

DemiMarie:
> Absolutely no m4 at all, anywhere.
> 
> Mind you, I have no thoughts on how we get there from here at the moment.

The first idea that comes to mind is to use a skeleton file that is compiled to C by a Perl script.

PR #413, opened 2019-01-08
build: use AC_SYS_LARGEFILE in configure.ac

ldv-alt:
This enables LFS on those of 32-bit systems that do not enable it
by default.

As result, stat() call in src/main.c:check_options() will not fail
on filesystems with 64-bit inode numbers.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,6 +29,7 @@ AC_PREREQ([2.60])
 AC_CONFIG_SRCDIR([src/scan.l])
 AC_CONFIG_AUX_DIR([build-aux])
 AC_USE_SYSTEM_EXTENSIONS
+AC_SYS_LARGEFILE
 LT_INIT
 AM_INIT_AUTOMAKE([1.15 -Wno-portability foreign std-options dist-lzip parallel-tests subdir-objects])
 AC_CONFIG_HEADER([src/config.h])


Explorer09:
I wonder, how can this Large File Support be useful in Flex? Was there a real use case for this, or are you adding LFS just for the sake of it?

ldv-alt:
There is the stat() invocation mentioned above that actually fails with EOVERFLOW on filesystems with 64-bit inode numbers.

Explorer09:
@ldv-alt I'm not convinced. Why would a 32-bit operating system need to care about a "filesystem with 64-bit inode numbers"?

ldv-alt:
If your OS is 32-bit, it doesn't mean that all filesystems it has to deal with have 32-bit inode numbers, just think about network filesystems - most of them have 64-bit inode numbers nowadays.

I'm sorry but what I'm talking about is very basic Linux knowledge, please read the paragraph about EOVERFLOW in http://man7.org/linux/man-pages/man2/stat.2.html

Explorer09:
@ldv-alt Sorry, but I personally treat that as a niche use case. I mean, you may add large file support for your build of Flex, but why should this upstream care about it? Most users use 64-bit operating systems nowadays (and 32-bit is left for legacy hardware or applications). Wouldn't the long-term solution for your case be switching to 64-bit OS, and LFS serves only as a temporary solution?

ldv-alt:
I'm the maintainer of flex package in a GNU/Linux distribution, we have been providing 32-bit OSes along with 64-bit OSes for many years, and some architectures will stay 32-bit and actual for many years to come, far beyond 2038.

I'm happy to report upstream a bug and suggest a fix for it I've already applied in my tree, this is a kind of bugs users report as "flex does not work".

Sadly, the upstream of flex doesn't push any commits to the upstream flex tree any longer, so this pull request seems to be a waste of time.

Explorer09:
@ldv-alt To clarify, I'm not the maintainer of Flex. I'm just one of the voluntary contributors here, and just wonder about the real intention of this PR (which you didn't mention until I asked that).

My thought is, if LFS is so useful that it's worth enabling it universally, then Autoconf should have been made it the default already, and it wouldn't be Flex that need to bother with it.

In other words, compatibity with a network filesystem shouldn't be a priority for Flex or many user application except those that are explicitly designed to work with large files.
You have the `-D_FILE_OFFSET_BITS=64` workaround (that can apply distro-wide) already, and it's better to suggest to Autoconf to make that the default, instead of pushing responsibilities for enabling that to many of us userspace programs.

Explorer09:
P.S. Many of the Flex code and internal structures are not yet large file aware. If someone can help evaluate or improve the code so that it becomes large file ready then it would be more chance for us to finally use AC_SYS_LARGEFILE. Enabling that without a code review (to ensure that nothing breaks with large file) seems for me a bad idea.

ldv-alt:
While it's a good idea to enable -D_FILE_OFFSET_BITS=64 in libc by default for all new 32-bit architectures, this cannot be done safely enough on traditional 32-bit architectures because it's an ABI change and some not quite portable software may be unprepared for this change.
For the same reason this is not going to happen in Autoconf, too.

It doesn't mean that the change of default e.g. in glibc will not happen in the future, it means that software cannot assume that this change will happen soon.
See also https://sourceware.org/ml/libc-alpha/2019-01/msg00178.html

I've verified that flex does not regress when built in -D_FILE_OFFSET_BITS=64 mode.

westes:
The OS and filesystem might have different origins.

On Friday, 11 January 2019,  4:08 pm -0800, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> @ldv-alt I'm not convinced. Why would a 32-bit operating system need to care about a "filesystem with 64-bit inode numbers"?
> 
> -- 
> https://github.com/westes/flex/pull/413#issuecomment-453696875

-- 
Will Estes
westes575@gmail.com

Explorer09:
@ldv-alt I would be happy if you could make a more thorough analysis to the code to ensure that Flex is truly large file ready (use proper `off_t` type when it's appropriate, and not assume the whole file content can be loaded into memory (since `sizeof(off_t) > sizeof(size_t)`), etc.) A compile time regress check is not enough, you know. (There's a lot of mess in Flex code when I last looked at it, and I would like Flex's code be modernized, if I have time to help do so.)

@westes My worries is about the compatibility with rest of the tools in the toolchain. Flex depends on M4, and the Flex-generated code would go to the C compiler for turning to machine code. All of this would need to support large file, or large file would not work. That's why I was skeptical with this PR at first.

westes:
Agreed on all points, honestly. Yes, large file support needs to work all up the chain and yes, flex needs cleanup/modernization.

On Saturday, 12 January 2019,  5:47 am -0800, "Kang-Che Sung (宋岡哲)" <notifications@github.com> wrote:

> @ldv-alt I would be happy if you could make a more thorough analysis to the code to ensure that Flex is truly large file ready (use proper `off_t` type when it's appropriate, and not assume the whole file content can be loaded into memory (since `sizeof(off_t) > sizeof(size_t)`), etc.) A compile time regress check is not enough, you know. (There's a lot of mess in Flex code when I last looked at it, and I would like Flex's code be modernized, if I have time to help do so.)
> 
> @westes My worries is about the compatibility with rest of the tools in the toolchain. Flex depends on M4, and the Flex-generated code would go to the C compiler for turning to machine code. All of this would need to support large file, or large file would not work. That's why I was skeptical with this PR at first.
> 
> -- 
> https://github.com/westes/flex/pull/413#issuecomment-453748928

-- 
Will Estes
westes575@gmail.com

JL2210:
Not sure if this has already been resolved, but 64-bit filesystems and 64-bit OSs are completely different. Quite a few filesystems are 64-bit nowadays, including ext4, NTFS, exFAT, and many others. You don't need to have a 64-bit operating system to access or use a 64-bit filesystem. I can access my 64-bit ext4 filesystem just fine from my 32-bit Gentoo computer. I have noticed Flex not working as well on it, however (issues finding files).

Another note about this: Almost all software has large file support by using `-D_LARGEFILE64_SOURCE`, which is set by `AC_SYS_LARGEFILE` in `configure.ac`.

All in all, I'd say this is a small change that will most likely fix more issues than it causes.
Just my two cents.

Explorer09:
@JL2210 I'm personally not against Large File Support in flex. However, do remember that a proper large file support is more than tuning the compile flags. And a PR that just adds a comfigure flag without code examination on the back makes me suspicious about its impact. You probably won't build a scanner that's gigabytes large, so Flex hasn't been written with large file awareness in the first place. What I wished is a thorough code analysis so that nothing could break when we add the compile flags.

JL2210:
I get that, and I agree with you on the fact that nobody's going to build a scanner that's more than 4 gigabytes, but please take note of this from the original PR description:
> As result, stat() call in src/main.c:check_options() will not fail
on filesystems with 64-bit inode numbers.

I think that this was the main thing that the PR was supposed to fix. Anyway, building from source after running `autoreconf -vif` with this patch applied looks fine (latest binutils builds and works).

JL2210:
64-bit Testsuite results with this patch applied:
```
Note: configured with `./configure --prefix=/usr/local --libdir=/usr/local/lib64 --enable-shared --disable-static`
============================================================================
Testsuite summary for the fast lexical analyser generator 2.6.4
============================================================================
# TOTAL: 114
# PASS:  114
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

32-bit Testsuite results with this patch applied:
```
Note: configured with `./configure --prefix=/usr/local --libdir=/usr/local/lib --enable-shared --disable-static`
============================================================================
Testsuite summary for the fast lexical analyser generator 2.6.4
============================================================================
# TOTAL: 114
# PASS:  114
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

Please note that @ldv-alt has also tested:
> I've verified that flex does not regress when built in -D_FILE_OFFSET_BITS=64 mode.

Explorer09:
@JL2210 The flex testsuite is not often maintained, unfortunately. The suite doesn't have any stress tests about large sizes, for example.

As I said before, I consider building a 32-bit flex program to work with a 64-bit filesystem/inode a niche case, and a better solution would be to switch to a 64-bit OS entirely. If there's someone responsible for the incompatibility, it would be the OS or Autoconf, and we shouldn't be bothered to fix anything to make the mentioned case work.

JL2210:
Like I said before, 64-bit filesystems are *totally* different than 64-bit operating systems. They're almost entirely unrelated, other than the (unfortunate) fact that large file support is not enabled on 32-bit operating systems. I don't believe this was meant to enable large file support as much as enabling using Flex on newer filesystems (listed earlier). I imagine that some people also don't have a choice as to whether they can use a 64-bit operating system or not. In any case, Flex continues to work with this enabled, so is there any reason at this point not to include it?

Explorer09:
@JL2210 In other words, the decision of pushing AC_SYS_LARGEFILE without making flex to be truly large file aware is a political one. And I hate politics being pushed here.

Is there any modern hardware not able to run a 64-bit OS (excluding VMs)? And how can it be a norm for users with a 32-bit OS expect they can connect to a 64-bit filesystem and work with it flawlessly (it's users' decision to keep a 32-bit hardware)? The patch needs these two assumptions to make sense, neither of which had I agreed with. The rest is politics.

JL2210:
Some Chromebooks either cannot run a 32-bit OS or do not because of hardware limitations. As Chromebooks cannot install alternate operating systems, that is not user choice. Along with that, several physical filesystems that are used on Chromebooks, eCryptFS and ext4, are 64-bit and are also not user choice. Yes, I know that that's probably not a viable use-case, but [Chromebrew](https://github.com/skycocker/chromebrew) does require it.
Once again, the bitness of operating systems and the bitness of filesystems are unrelated, so users of a 32-bit OS should be able to use a 64-bit filesystem with no problem.
I don't think it would be right to force a user to have to use a 64-bit OS if they want to use a 64-bit filesystem, when it could be fixed simply by adding a macro to `configure.ac`.

Explorer09:
@JL2210 
The very first question of your mentioned case is: Why should _we_ bother with the Chromebook builds? No matter which package manager you use, package managers should be able to maintain their local patches, don't they? Why is that our upstream matter to patch things for them, especially when their case is uncommon already?

JL2210:
So are you dropping support for 64-bit filesystems on 32-bit operating systems? It's just one m4 macro, and it doesn't break anything.

Explorer09:
@JL2210 I didn't say that. What I said is there's no thorough code analysis for any potential breakage when LFS is enabled in Flex, and adding a simple m4 macro without that would give a false impression that Flex is large file ready which in fact it doesn't.

And I suspect the trivial addition of the macro could have political reason that needs to be revealed and discussed first (just because it's too trivial). If the PR was about real large file support in flex and it proves it works, then I'm quite happy to see it be merged in. But this PR isn't about that.

ldv-alt:
The fact that I submitted this means that of course I checked all flex code paths.
For example, the sole purpose of stat() call in flex is to check its return value and st_mode field.
If you do a trivial "git grep fseek", you'll see that flex is not prepared to handle large files even on 64-bit filesystems.

ldv-alt:
The fact that I submitted this means that of course I checked all flex code paths.
For example, the sole purpose of stat() call in flex is to check its return value and st_mode field.
If you do a trivial "git grep fseek", you'll see that flex is not prepared to handle large files even in 64-bit operating systems.

JL2210:
I'd be happy to help send a patch to enable large file support in Flex, but this is probably the first step in being able to do that.

Explorer09:
@JL2210 @ldv-alt 
I read the now commit message just now, unfortunately it just explain the obvious of what `-D_FILE_OFFSET_BITS=64` would do. But one question remains unanswered: Is there no alternative from using `stat()` for the context that's needed in flex code?

The other fopen() set of calls seems to be not affected by bit depth of the inode. Please prove me wrong.

Explorer09:
@ldv-alt Hello. Please try #429. If your motivation of pushing AC_SYS_LARGEFILE to flex is only to address the 64-bit inode number problem, then it would be better to just remove the stat() system call.

westes:
While #429 needs to have a file conflict resolved, I agree it will be better to go that route.

PR #429, opened 2019-10-30
Remove manual m4 search code. No more stat() dependency.

Explorer09:
This reverts commit c34590c4bf067d4e749e5298a8ce9f90072d6332
and all search code patches that followed it.

The execvp() would do the $PATH searching of m4 for flex. There's no
need to duplicate the effort.

Another reason for the removal is to remove dependency on stat()
system call, which would cause a problem in a 32-bit OS accessing a
filesystem with 64-bit inode number (see #413).


Explorer09:
Rebased onto current master.

Issue #377, opened 2018-08-15
GCC 7.3.0 warns about emitted code: potential null pointer dereference


akimd:
I installed the following patch in Bison:

```diff
commit ef98967ada3c1cd48c177d7349e65a709bb49b97
Author: Akim Demaille <akim.demaille@gmail.com>
Date:   Wed Aug 15 13:49:01 2018 +0200

    build: work around GCC warnings on Flex code
    
    With GCC 7.3.0 and Flex 2.6.4, we get warnings on all the generated
    scanners:
    
        examples/calc++/calc++-scanner.cc: In function 'void yyrestart(FILE*)':
        examples/calc++/calc++-scanner.cc:1611:20: error: potential null pointer dereference [-Werror=null-dereference]
         /* %endif */
          ~~~~~~~~~~~       ^
        examples/calc++/calc++-scanner.cc:1607:19: error: potential null pointer dereference [-Werror=null-dereference]
         /* %if-c-only */
          ~~~~~~~~~~~~~~~  ^
        examples/calc++/calc++-scanner.cc:1611:20: error: potential null pointer dereference [-Werror=null-dereference]
         /* %endif */
          ~~~~~~~~~~~       ^
        examples/calc++/calc++-scanner.cc:1607:19: error: potential null pointer dereference [-Werror=null-dereference]
         /* %if-c-only */
          ~~~~~~~~~~~~~~~  ^
        cc1plus: all warnings being treated as errors
    
    Obviously the lines are incorrect, and the warnings are emitted twice.
    Still, let's get rid of these warnings.
    
    * doc/bison.texi, src/flex-scanner.h: Disable these warnings in code
    generated by Flex.

diff --git a/doc/bison.texi b/doc/bison.texi
index 9f16f730..2882506b 100644
--- a/doc/bison.texi
+++ b/doc/bison.texi
@@ -11399,6 +11399,12 @@ parser's to get the set of defined tokens.
 # undef yywrap
 # define yywrap() 1
 
+// Pacify warnings in yy_init_buffer (observed with Flex 2.6.4)
+// and GCC 7.3.0.
+#if defined __GNUC__ && 7 <= __GNUC__
+# pragma GCC diagnostic ignored "-Wnull-dereference"
+#endif
+
 // The location of the current token.
 static yy::location loc;
 %@}
diff --git a/src/flex-scanner.h b/src/flex-scanner.h
index 4dec7e1f..e09a1c04 100644
--- a/src/flex-scanner.h
[...]

PR #431, opened 2019-11-04
scanner: Fix compiler confusion in yy_init_buffer()

Explorer09:
When complied with 'gcc -O3', the yy_init_buffer call can confuse gcc,
thinking the 'b' pointer may be NULL. (gcc would warn that if
'-Wnull-dereference' is used). Fix the confusion by never pass a NULL
constant to the function.

Fixes: #377

### DIFF
--- a/src/c99-flex.skl
+++ b/src/c99-flex.skl
@@ -854,7 +854,7 @@ void yyrestart(FILE * input_file, yyscan_t yyscanner)
 			yy_create_buffer( yyscanner->yyin_r, YY_BUF_SIZE, yyscanner);
 	}
 
-	yy_init_buffer( yy_current_buffer(yyscanner), input_file, yyscanner);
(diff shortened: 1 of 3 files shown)


akimd:
LGTM.
FWIW, I would use `YY_CURRENT_BUFFER_LVALUE` in the definition of `YY_CURRENT_BUFFER` to clarify the connection between both of them.

Explorer09:
@akimd I cannot say if changing the definition of `YY_CURRENT_BUFFER` would work without causing any regression.

akimd:
I meant
``` C
#define YY_CURRENT_BUFFER ( YY_G(yy_buffer_stack) \
                          ? YY_CURRENT_BUFFER_LVALUE \
                          : NULL)

```

Explorer09:
@akimd The two macros are defined close together. What you suggested is a cosmetic change. I won't bother.

Explorer09:
Rebased onto current master.

PR #449, opened 2020-06-24
Constify local 'slash'

kevans91:
slash is used for the return value of `strrchr` on M4, which is by default
a macro expanding to a string literal.

While `strrchr` returns a non-const `char *` on contract, we should not be
modifying the return value of it based on M4. Some versions of GCC will
in-fact warn about this, even though the return value isn't being mutated
right now.

This has the side-effect of making it clear that *slash should not be
mutated, whether it is currently or not.


kevans91:
This appears to be overcome by events...

PR #456, opened 2020-09-12
WIP: Make lex retargetable to languages other than C

eric-s-raymond:
Not finished yet, but there's enough here for you to see that the architecture works.  The tip version passes all tests, so this could be merged now.

This patch series uses two different strategies for isolating C-specific assumptions:

1. Abstract various things we might want to generate in C to a method table that knows (for example) what the start and end of array initializers need to look like, and how to emit constant declarations.

2. Move conditional generation of procedural code from being done by C functions in gen.c to being done by conditionalized macros in the m4 skeleton.

Translating gen_find_action turned out to be quite hairy, which is why I haven't finished this patch series yet. Most of its guts have been replaced by a new M4_REJECT_FIND_ACTION macro.  It is possible that by the time I'm done that macro will replace the entire function, and then the macro body will get inlined at its former call site.  Which would be good, as it would make generated parser code far easier to read.


eric-s-raymond:
I'm withdrawing this pull request.  I've decided to redo the patch set; as it is now, it is overly long and complicated because I had to change strategies in the middle, and it is too difficult to verify that the changes are correct.

New pull request to follow shortly.

PR #457, opened 2020-09-21
Retargeting

eric-s-raymond:
Here's the cleaner pull request.

The comment of the last commit describes how to audit to be sure the code generated by the tests is stable.  So far, the only change is that #line directives are gone (this makes comparing trivial whitespace changes easier).

This patch set isn't caught up with my old go branch yet.  I need to redo most of the the work on gen.c and the skeleton file in easily auditable chunks. On the other hand, this patch series has provable correctness; it can land on master immediately, and should.


westes:
duplicate of #460

PR #461, opened 2020-09-28
Fix build on macOS by using GNU sed in mkskel.sh

arichardson:
The sed commandline used in mkskel.sh does not work correctly when using
the macos sed. Use the version of GNU sed found by configure instead to
generate a valid .c file.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,6 +55,7 @@ AM_PROG_CC_C_O
 AC_PROG_LN_S
 AC_PROG_AWK
 AC_PROG_INSTALL
+AC_PROG_SED
 
 pkgconfigdir=${libdir}/pkgconfig
 AC_SUBST(pkgconfigdir)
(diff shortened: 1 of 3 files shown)


Explorer09:
What is the problem with macOS sed? sed is a POSIX standard utility, and its syntax is supposed to be portable. It's better to fix the particular portability bug with macOS sed than to force another sed implementation as a workaround.

arichardson:
mkskel.sh does not work with POSIX sed. First of all the multiple statements need to be changed to multiple -e flags, but even then the generated output does not compile. I assume some of the sed expressions used  rely on GNU extensions.
```sh
sed 's/4_/a4_/g
s/m4preproc_/m4_/g
' "$srcdir/flex.skl" |
"$m4" -P -I "$srcdir" "-DFLEX_MAJOR_VERSION=$1" \
   "-DFLEX_MINOR_VERSION=$2" \
   "-DFLEX_SUBMINOR_VERSION=$3" |
sed '/^%#/d
s/m4_/m4preproc_/g
s/a4_/4_/g
s/[\\"]/\\&/g
s/[^\r]*/  "&",/'
```

Explorer09:
Without the exact output and the error log, I cannot track exactly which problem you are facing in particular.

* Multiple sed statements separated by line breaks are legal. No need to use multiple `-e` flags for that.
* The only portability problem I've seen so far is the `\r` escape sequence. I've addressed it in #353. Try the patch and see if it builds.

arichardson:
Thanks, I can confirm it works with #353

PR #462, opened 2020-09-28
Fix -Wundef warnings by using #ifdef instead of #if

arichardson:
This also brings back the check for __func__ that was deleted in commit
9b41a091bcb2fb9bd2b6db79d06a0c976d1a5663


arichardson:
ping?

westes:
Now that #489 is merged in, I will have a look at this.

westes:
This is now on master, albeit ever so slightly altered from your pr.

PR #466, opened 2020-11-02
Add enumsc and enumscstr.

phidebian:
Add enumsc and enumscstr, an enum for start cond instead of define and the optional enum to string array for easier debug printout.


phidebian:
I am writting a lex scanner (actually many) with a good number of SC, beeing able to printout the SC in text is a good help in the debug print out. the cost of change is modest.

Note I did this quick'n'dirty, I didn't check if this exit already in other clones.....

Cheers.

westes:

Explorer09:
Flex allows `bool` type for the `enumsc` and `enumscstr` flags. Please use that instead of `int`. It helps readability.

PR #468, opened 2020-12-08
flex: Fix typo in noline option check

mobbitster:
If we set --noline we should not print line directives.
But setting --noline means gen_line_dirs is false.

Signed-off-by: Oleksiy Obitotskyy <oobitots@cisco.com>


GitMensch:
I think this was superseded by https://github.com/westes/flex/commit/64cf0328063cd97dec09536a8f2bb35ce7350ffe#diff-730257e2facf9dd0d1962d47c08b1d360bd867f906c11dbf50ce96c50d4abd9c and should be closed.

PR #470, opened 2020-12-15
src/gen.c: Remove dead code from do_indent()

Sawuare:
In `do_indent()`, since `i` is a multiple of 8 and is decremented by 8 `while (i >= 8)`, `while (i > 0)` is never true.


Sawuare:
Closing this because the function was removed in commit 8521281aef33b47df7e372d16a495cec7f9a7e62 and my fork was outdated.

PR #474, opened 2021-01-24
Emit code from the skeleton or skeleton macros.

Mightyjo:
Hello!

I finally finished the fork I started back in 2015 to move all the C code emissions into the skeleton. This will make maintenance of the built scanner simpler by keeping it (almost) all in one place instead of spread out among the flex sources. 

The few emissions that have to come from flex are done through skeleton macros, now. This is to enable my next project - multi-language support. I'm going to start by improving -S option support and then split the C and C++ scanners apart as a first example. 

The squashed log message for the PR follows:

Move all C language emissions to the skeleton and translate the
options logic to m4.

Emit top_buf and userdef_buf into m4 vars.

Add emission of top_buf and userdef_buf m4 var values to flex.skl.

The %top block and user-defined preprocessor definitions are benign
chunks of target code that can be emitted wholesale into the scanner.
However, their placement in the scanner source should be determined by
the skeleton rather than hardcoded in the main.c:check_options() function.
Now it is.

Change emission of YY_INT_ALIGNED to definition of an m4 variable.
Add emission of new m4 variable to flex.skl.

The optimization intended by YY_INT_ALIGNED is tied to the target
language.  The skeleton should be responsible for whether and where
to emit the type definition, even as the user is responsible for
determining whether it is wanted.

Add macros defining language-specific table structure to flex.skl.
Refactor dfa.c and misc.c methods used by ntod() to use the new table macros.
Replace obsolete yyconst uses with const.

Move CLI and %option switches from main.c to flex.skl.
Document the M4_ macros that implement the switches in flex.skl for the
benefit of skeleton authors.

Define a flag when backing-up states are used.
Transliterate gen_backing_up() and gen_bu_action() into macros.

Add variadic buf_printns like buf_prints to handle macro emissions with 2
or more parameters.

Add out_m4_define_hex that prints its numeric argument in hex format.

Add skeleton macros for emitting table elements and references.

Start using YYDMAP_ENTRY macros for emitting yydmap table entries.

Remove dead code.

### DIFF
--- a/src/buf.c
+++ b/src/buf.c
@@ -71,13 +71,41 @@ struct Buf *buf_print_strings(struct Buf * buf, FILE* out)
 struct Buf *buf_prints (struct Buf *buf, const char *fmt, const char *s)
 {
 	char   *t;
-        size_t tsz;
+    size_t tsz;
 
 	tsz = strlen(fmt) + strlen(s) + 1;
 	t = malloc(tsz);
 	if (!t)
 	    flexfatal (_("Allocation of buffer to print string failed"));
 	snprintf (t, tsz, fmt, s);
+	buf = buf_strappend (buf, t);
+	free(t);
+	return buf;
+}
+
+/* Append a "%s %s ..." formatted string to a string buffer */
+struct Buf *buf_printns (struct Buf *buf, const char *fmt, const int count, ...)
+{
+	char   *t;
+    size_t tsz;
+	va_list ap;
+	const char* s;
+	
+	va_start(ap, count);
+	tsz = strlen(fmt);
+	for(int i=0; i<count; ++i) {
+	    s = va_arg(ap, const char*);
+		tsz += strlen(s);
+	}
+	va_end(ap);
+	tsz += 1;
+	t = malloc(tsz);
+	if (!t)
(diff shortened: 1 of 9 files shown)


Mightyjo:
See the skeleton_pr branch on [TravisCI](https://travis-ci.com/github/Mightyjo/flex).

westes:

Mightyjo:
On it now

Mightyjo:
Took a while to rebase this after ESR's change set. There's not much significant left: Using of the new M4_YY_EOF_STATE_CASE_ARM hook in parse.y is worth keeping. Otherwise it's just some comments and whitespace.

I'll squash this back down and simplify the commit messages.

My next step is to write a code emission backend that will avoid the extra M4 hooks recently introduced. I can base that on this branch and let everyone watch or start another PR when it's further along.  Preference?

westes:
I think starting a new branch after I get this merged is the best/easiest way; it will leave you free to work and unblocks things as I clean out the backlog of pr's.

Mightyjo:
Sounds good. I've started and stopped a new branch a few times. Think I'm on a working track with it now.

Issue #469, opened 2020-12-14
Flex causes memory corruption and eventually causes program to crash


satya-das:
Following program crashes because of memory corruption:

```c++
%{
#include <fstream>
#include <vector>

#define ECHO
%}

%option noyywrap

ID	 [_a-zA-Z]+[_0-9a-zA-Z]*

BBC1 ([^(]*|"("[^(]*")")

/**
 * Balanced Bracketed Content.
 * May not contain bracket, but if it does then it is balanced.
 */
BBC ({BBC1}*|"("{BBC1}*")")

FPD "("{BBC}*")"

%%

{ID}/{FPD}\{ {
  return 1234;
}

%%

std::vector<char> readFile(const std::string& filename)
{
  std::vector<char> contents;
  std::ifstream     in(filename, std::ios::in | std::ios::binary);
  if (in)
  {
    in.seekg(0, std::ios::end);
    size_t size = static_cast<size_t>(in.tellg());
    contents.resize(size + 2);
    in.seekg(0, std::ios::beg);
    in.read(contents.data(), size);
    in.close();

    contents[size + 0] = '\0';
    contents[size + 1] = '\0';
  }
  return contents;
}

int main(int argc, char** argv)
{
  auto stm = readFile( "drawable.h");
  yy_scan_buffer(stm.data(), stm.size());

  yylex();

  return 0; // All went well.
}
```
This program crashes when `drawable.h` is some particular content which is basically a C++ file. That can be fetched from here: https://github.com/satya-das/cppparser/blob/master/test/e2e/test_input/ObjectArxHeaders/drawable.h

Sorry, I could not report this issue in any other way, it was the only way I could simplify the program I am writing.
To reproduce the crash. First download the [drawable.h](https://github.com/satya-das/cppparser/blob/master/test/e2e/test_input/ObjectArxHeaders/drawable.h)  in the same folder as above program, assuming the name of file of above program is `parser.l`:
```sh
flex -oparser.cpp parser.l
g++ -oparser parser.cpp

./parser
```
Then a crash will be seen (most probably). If not then use valgrind to get memory report:
```sh
valgrind ./parser
```
On my machine the valgrind report is:
```
==13217== Memcheck, a memory error detector
==13217== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==13217== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==13217== Command: ./parser
==13217== 
==13217== Invalid write of size 4
[...]

PR #475, opened 2021-01-27
Compute length of yy_state_buf correctly 

Mightyjo:
Corrects issue #469 as reported. 
[TravisCI build report](https://travis-ci.com/github/Mightyjo/flex/builds/214741288) is clean.

There might still be a problem if a user does a loop:
```
{
  yy_scan_buffer(); 
  yylex(); 
}
```
Couldn't see whether it's correct to invalidate yy_state_buffer when yy_scan_buffer() is called.

This issue doesn't crop up when scanning from a stream. In that case the scanner aborts when it attempts to grow the input buffer.


Mightyjo:
Was able to confirm that the loop I described above will cause corruption if the later instance of yy_scan_buffer introduces an even longer input.

Working on a solution.

Mightyjo:
Solved.

Added tracking of the reject state buffer size to the standard C, reentrant, and C++ scanners.

Added a check to yy_switch_to_buffer that ensures enough space is available in the reject state buffer. That gets called whenever the buffer changes due to YY_INPUT, actions, or user intervention and that's all the cases. 

Added two tests that were able to trigger the overflow described in #469 when the input buffer exceeds YY_BUF_SIZE. These will fail if we have a reversion.

smcpeak:
I'm not a `flex` maintainer, but I ran into this bug myself, then found #469, so looked at your proposed fix.  I think there are a few issues.

----

First, the proposed fix is not consistent about whether `yy_state_buf_max` is a count of bytes or elements.  In particular, in `flex.skl`, you have both:

```
YY_G(yy_state_buf) = (yy_state_type *)yyalloc(YY_STATE_BUF_SIZE  M4_YY_CALL_LAST_ARG);
```

and

```
YY_G(yy_state_buf_max) = YY_STATE_BUF_SIZE;
```

implying it is a byte count, but also:

```
YY_G(yy_state_buf) = (yy_state_type *)yyalloc( (YY_CURRENT_BUFFER_LVALUE->yy_buf_size + 2) * sizeof(yy_state_type)  M4_YY_CALL_LAST_ARG);
```

and

```
YY_G(yy_state_buf_max) = (YY_CURRENT_BUFFER_LVALUE->yy_buf_size + 2);
```

and

```
yy_state_buf = new yy_state_type[YY_STATE_BUF_SIZE];
yy_state_buf_max = YY_STATE_BUF_SIZE;
```

implying it is an element count.

I suggest adding a comment by its declaration that says exactly what it is, then changing the code to match.  Based on the rest of the code, it looks like it is meant to be an element count.

----

Second, this test:

```
if ( YY_G(yy_state_buf_max) < YY_CURRENT_BUFFER_LVALUE->yy_buf_size ) {
```

is missing a +2.  `yy_buf_size` is the number of bytes in the input buffer *minus two*, and it appears `yy_state_buf_max` is meant to include that +2, but the test fails to account for that.

----

Third, the +2 on all allocations of `yy_state_buf` should actually be a +3.  The old +2 is a pre-existing bug in `flex` (going back to at least `flex-2.5.4`) that ideally should be fixed at the same time.  The reason for +3 is the buffer contains the start state as well as the state after every byte.

----

Fourth, the line:

```
new_size = YY_G(yy_state_buf_max) + YY_CURRENT_BUFFER_LVALUE->yy_buf_size;
```

does not make much sense to me.  The new size is the sum of the old size and the required size?  That allocates more than needed.  Also the +2 (+3) is missing.  I think it should simply be:

```
new_size = YY_CURRENT_BUFFER_LVALUE->yy_buf_size + 3;
```

----

I independently found and fixed this bug in [my own fork of flex](https://github.com/smcpeak/smflex).  I have a [writeup of the bug](https://raw.githubusercontent.com/smcpeak/smflex/smflex/historical/yy-state-buf-bug.txt) and [my fix](https://github.com/smcpeak/smflex/commit/9411f5cf492b4398cd54bc9755a42003d4473c11).  You (and anyone else) are welcome to use that material, but be aware the diff will not apply cleanly to `flex` because of the many unrelated changes I have made (but it should be straightforward to make the same changes manually; the core data structures are the same, even though some fields have been renamed).

Also be aware I haven't tested any of the changes I proposed above on `flex`.  My suggestions are based on what I did (and tested) in my fork.

Mightyjo:
Thank you for checking my work! You're right, I confused myself over whether I was working with a count of elements or bytes. 

I have ensured that YY_STATE_BUF_SIZE and yy_state_buf_max are both element counts.
I added a constant YY_STATE_BUF_EXTRA_SPACE representing the 3 additional states in the state buf vs the char buf.

Thanks for your notes on the third additional state, by the way!


> ```
> if ( YY_G(yy_state_buf_max) < YY_CURRENT_BUFFER_LVALUE->yy_buf_size ) {
> ```
> 
> is missing a +2. `yy_buf_size` is the number of bytes in the input buffer _minus two_, and it appears `yy_state_buf_max` is meant to include that +2, but the test fails to account for that.

I changed this to 
 ```
 if ( YY_G(yy_state_buf_max) < (YY_CURRENT_BUFFER_LVALUE->yy_buf_size + YY_STATE_BUF_EXTRA_SPACE ) ) {
 ```
on the assumption that the +3 v. +2 applied here. It passes my regression tests, but please let me know if I've missed a subtlety.


> Fourth, the line:
> 
> ```
> new_size = YY_G(yy_state_buf_max) + YY_CURRENT_BUFFER_LVALUE->yy_buf_size;
> ```
> 
> does not make much sense to me. The new size is the sum of the old size and the required size? That allocates more than needed. Also the +2 (+3) is missing. I think it should simply be:
> 
> ```
> new_size = YY_CURRENT_BUFFER_LVALUE->yy_buf_size + 3;
> ```

I didn't like how much memory I was using there either. I've switched to your suggested allocation. It works for the regression test in my PR. 

A little about why I wrote it that way: 
If I understand correctly, the reject buffer carries over from one input buffer to the next so you can implement things like include directives. Normally, reject states get freed regularly and we never grow to the max. I was thinking of a perverse test case involving an include directive within a balanced brace as in the regression tests. We don't have the memory management facilities to tell how large the combined buffer would be, so it seemed safer to add enough state space for the entire new buffer rather than just the excess portion.

Code push incoming.

src/flex.skl
```diff
@@ -1994,20 +1996,22 @@ void yyFlexLexer::yyrestart( std::istream* input_file )
...
-          else {
-            YY_G(yy_state_buf) = new_state_buf;
-            YY_G(yy_state_buf_max) = new_size;
-          }
-        }
+		/* Ensure the reject state buffer is large enough.
+		 */
+		if ( YY_G(yy_state_buf_max) < (YY_CURRENT_BUFFER_LVALUE->yy_buf_size + YY_STATE_BUF_EXTRA_SPACE) ) {
+			/* yy_state_buf_max already contains YY_STATE_BUF_EXTRA_SPACE */
+			/* new_size = YY_G(yy_state_buf_max) + YY_CURRENT_BUFFER_LVALUE->yy_buf_size; */
```
smcpeak, line 2003: This line of commented code looks like it was meant to be removed.

smcpeak:
The new changes look good to me.  I haven't tested them, but they seem to address the problems I saw before.

Regarding the possibility of `yy_state_buf` spanning buffers, I do not think that is possible when the API is used as intended.  At the start of every attempt to match a token, `yy_state_ptr` is reset to the beginning of the buffer.  So if a user action sets up a new buffer and then lets matching continue as normal, `yy_state_ptr` will be reset, so `yy_state_buf` only needs to be big enough for the new buffer.

That said, I think I see a code path that would break this: if an action sets up a new buffer and then calls `REJECT`, that necessarily bypasses the usual reset of `yy_state_ptr`.  I don't know why a scanner would do that, and it seems contrary to the API intent, but it is a potential risk.  Although this specifically isn't the reason, I'll note that, in my fork, every dereference of `yy_state_ptr` is protected by an assertion that it is within bounds.

src/flex.skl
Mightyjo: Yep, thanks for catching that. Fixed it.

Mightyjo:
I'll be satisfied that we're doing the Right Thing until I see a test that breaks, then. At the moment, it looks like all the uses of yy_state_ptr are emitted from compiled code instead of the skeleton. That's being fixed, but in the meantime tracing and fixing hypotheticals like the ones we're thinking up is absurdly difficult. Easier with a test in the harness, at least.

westes:

Mightyjo:
This is going to take a minute. The new skeletons are different enough that I'm having to work out three different solutions.

Mightyjo:
Hm. I can't find the machinery to handle reject states in non-reentrant scanners. Was that ever supposed to work?

westes:
> Hm. I can't find the machinery to handle reject states in non-reentrant scanners. Was that ever supposed to work?

I think it was, but I hazily remember it being buggy or incomplete or something.

Mightyjo:
Cool. The test for it was working last week. I won't consider that a fluke. 

Git keeps partially merging all the new things in master into my PR branch so I'm having trouble tracing the old flow vs. the new. That's to say, it's 50/50 whether I'll reimplement this and submit a new PR.

Mightyjo:
Had to do evil things to my history. Don't merge this.

PR #480, opened 2021-03-27
Automate release

### LINKED ISSUE #337: Automate flex release
Currently, flex releases are manual. This takes time and is error prone.

We would like for release tags to trigger flex releases.

Such a system needs to:

- build the flex release archives.
- generate release notes based on the NEWS file.
- notify the flex-announce mailing list (NB it is ok if that email needs approval).
- increment the recorded current and previous version numbers of flex based on the release number from the release tag and what flex's version numbering is recorded as.

When we have that, it will be possible to release flex whenever an important fix or feature change appears on master or when sufficient time has passed since the previous release.

Mightyjo:
Migrate the CI workflow to Github Actions.
Automate the release workflow based on the Milestone system. Closes #337 
Automate artifact signing with detached GPG subkeys in Github repository secrets (not in the repo).


Mightyjo:
Forgot to mention, you can safely leave out commit 5e21333 if you want. It fixes a file permission oddity that was stopping distcheck on my WSL2 VM's tortured filesystem. 

If I'd thought about it, I'd have tested the build without that commit on Github Actions before sending the PR.

westes:
This is my next priority. Thank you for submiting this. I'll comb through and make sure it does what we want.

Mightyjo:
Coolness. There's a bit of setup for the release workflow with the pgp keys and your git secrets. The push and PR workflows are easy though.

GitMensch:
@westes wrote:

> This is my next priority. Thank you for submiting this. I'll comb through and make sure it does what we want.

That's quite cool, I suggest linking #337 to this PR to auto-close the issue on pull.

Even more cool as #337 was specified as "release is very likely to get done and published once this is available" and it is really hard to follow all the changes since the last release (more than 4 years ago) [and only very few people use a snapshot build - and a new release will also be taken up by the package maintainers].

Mightyjo:
I've been leaving the linking keyword out of my PR messages. Thanks for getting me to look that up!

westes (commented):
it's not just all about the milestones; sometimes patches come in via email or things can get commited directly to master.

How do you anticipate handling those cases?

I'm willing to enforce commit message formates if that helps make something automate-able.

But what are your thoughts?

.github/workflows/build.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches: [ master ]
+    tags-ignore:
+    - 'v*'
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-18.04
```
andy5995, line 14:
```suggestion
    strategy:
      fail-fast: false
      matrix:
        arch: [x64, x86]
        os: [ubuntu-18.04, ubuntu-latest]
        compiler: [gcc, clang]
    runs-on: ${{ matrix.os }}
```

I was just browsing through and thought I'd suggest some type of matrix... though not necessarily this one. :)

Mightyjo: Point taken. I'm focusing on the release machinery in this PR but more extensive matrix testing is a good thing. I'd like the matrix test workflow to run early in the integration pipeline so the release workflow can stay streamlined. I'm having to repeat a lot of build code between workflows, though. Ready to learn.

.github/workflows/milestone.yml
```diff
@@ -0,0 +1,196 @@
...
+      - name: Change version number
+        id: version
+        env:
+          tagname: ${{ steps.news.outputs.tagname }}
+        run: |
+          ver=$tagname
+          ver=${ver#v}
+          ver=${ver%-*}
+          echo "::set-output name=version::$ver"
+          sed "s/^\(AC_INIT.*generator\],\)\(.*\)\(,\[flex-help.*\)$/\1[$ver]\3/" $GITHUB_WORKSPACE/configure.ac >> $GITHUB_WORKSPACE/configure.ac.new
```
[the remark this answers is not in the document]
Mightyjo: I am but an egg. I forgot you could do that.

Mightyjo:
> it's not just all about the milestones; sometimes patches come in via email or things can get commited directly to master.
> 
> How do you anticipate handling those cases?

True enough. Starting with the milestone messages kept the demo manageable. I looked at using git2cl at first but the result seemed too fine-grained for the NEWS file.

> I'm willing to enforce commit message formates if that helps make something automate-able.

Some kind of commit message formatting could help. I've been practicing the semantic-release formats on my own work. Their repo has some tools and hooks that can help with adoption and automation.
> 
> But what are your thoughts?

I started off trying to summarize the commit log but that got out of hand quickly. At this point I'm imagining the milestones as a release backlog and the NEWS file as a snapshot of the finished board. I should be able to augmenting the milestone issues with some specially-formatted commits within the release window to cover emailed patches and similar cases.

I also preserve any text that's already in NEWS so we could add custom entries to the top before running the release workflow. This was this simplest thing that could possibly work.

Anywho, I expected this would need discussion. I'm all ears.

.github/workflows/build.yml
andy5995: >  I'm having to repeat a lot of build code between workflows, though. Ready to learn.

I'm not too great with Actions yet... your yml looks pretty good though, a lot more advanced than I have done.

Of course one way to prevent code repetition in these workflow files is using shell scripts, but I didn't see where you could do that in the use-cases you need.

One thing that might help trim a few lines is to combine some of the name/run lines. You don't need a "name" for every "run". Example:

```
      - run: ./autogen
      - run: ./configure
      - run: make
      - run: make check
      - run: make distcheck
```

or 

```
      - run:  |
         ./autogen
         ./configure
         make
         make check
         make distcheck
```

Or because that sequence of commands is used in build.yml and milestones.yml, put them in a shell script.

```
- run: sh .github/scripts/build.sh
```

andy5995: Oh, you could probably actually remove the `make check` since `make distcheck` does that automatically. 

Also, I find that using `make V=1` can be handy sometimes

Croydon:
I think this PR tries to do too much stuff at once and hence is staled. 
First introducing CI builds at all would be already a major step forward.
Release automation could be the next step afterwards.

Mightyjo:
Flex previously used TravisCI. The first bunch of commits in this PR
translate that pipeline to GitHub Actions as much as possible. I can split
that chunk out from automated releases if it'll get this moving.

Croydon:
I (obviously) can't speak for @westes, but I think this pull request would be easier to grasp/review when the migration of the CI from Travis to GitHub Actions would be split out in an own pull request

Mightyjo:
I've split the migration to Github Actions out to PR #523. I'll rebase or replace this PR once that one is dealt with.

Mightyjo:
Rebased the automated release actions after splitting out the codeQL and distcheck actions. Also switched to 'sed -i' as suggested. (I'll handle matrix builds in the build action a little later on.)

The outstanding question from before was [here](https://github.com/westes/flex/pull/480#issuecomment-881132888). 

> it's not just all about the milestones; sometimes patches come in via email or things can get commited directly to master.
> 
> How do you anticipate handling those cases?

My milestone_close.rb preserves the existing contents of the NEWS file, so it's feasible to add one-off messages to the top before starting the release. They'll just end up out of order with respect to the Issues. 

We could also do something completely different. I mentioned before that I tried several things before hitting on the idea of using Milestones and Issues as progress markers. The trouble is mostly in getting the commit messages for a new release from the GitHub API in a coherent order. I don't think I found the best solution, but I needed more brains to think about it before iterating down a hole.

> I'm willing to enforce commit message formats if that helps make something automate-able.

Yeah, this will help. I'm using loosely using [this](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format) format. If we did adopt that format, we could use [semantic-release](https://github.com/semantic-release/semantic-release) workflows to build the NEWS and/or ChangeLog.

Mightyjo:
> There's currently no CI generated tarball, is it?
> 
> If not then can we get it first (making the result of passing `make distcheck` available (and before running the tests just update the version string to contain the commit date, similar as it is done here)?

Like unsigned nightlies? That's doable. 

I haven't learned to save build products for use by other workflows so I've had to run 'make distcheck' in both the test and release flows. I welcome help on that.

Mightyjo:
> Some people use one key for nightly builds and another for releases, the first used for manual verification, the second also in distributions.

Totally supportable. Just looks like a second subordinate GPG key in this process.

> And ideally the same step would be executed and only extended for the milestone closed one - did @westes already decided that this would be the last step before an automated release? [I'm personally a fan of "tagging with a defined name = release"].

Yep, it's just a different trigger condition at the top of the script. @westes didn't pick the milestone close as the trigger. I guessed at his release process knowing it's easy to change the condition later.

> Note: I really like automated releases as good as it can be done, but _personally_ think a NEWS entry should consist of user-visible adjustments and deserve entries (from time to time, so it doesn't need to be done all directly before the release) written by a human. All other entries of this workflow seem very good to me.
> 
> Maybe the "generate NEWS stub" code could be just committed to the repository and be manually executed as "starter" for a human to check which entries to include in which way if that part is found to be necessary?

Same. Maybe the automated part should go in the ChangeLog? It's really hard to write/maintain two-part workflows, do I'm reluctant to include commit-pause-resume patterns to get human input. Anywhere we can pull the input from a ready made database is great. Places like the issues and milestone databases seemed perfect.

GitMensch:
@Croydon just re-asking for a review, because it was missing after the split of the merged CI actions.

Croydon:
> @Croydon just re-asking for a review, because it was missing after the split of the merged CI actions.

I'm not entirely sure why you ask me for a review. I don't have any repository rights here or something like that. So my opinion here is just as relevant or irrelevant as that of any other random person here.

But if you ask me, I'm wondering if the amount of automatization is worth the complexity. IMHO adjusting version numbers in the code and writing the changelog could happen manually, then you add a new git tag for the new version and the CI is creating a GitHub release, the official release downloads and uploads them to the GitHub release (and other targets if desired).

It would still automate the actual publishing part, but no complex magic how useful changelogs can be auto-generated and what changes need to be automatically created and pushed to git.

Mightyjo:
@Croydon fair points about complexity. 

I demoed everything I could find a way to automate. I don't expect us to keep it all. For example, I'm satisfied that the ChangeLog and the NEWS file serve different purposes. I can automate collection of the ChangeLog, but writing a good NEWS update is a whole research problem (or we could get a Large Language Model to do it).

Amusingly, if we start producing nightly builds it will be less complicated to update the version numbers because the automatic parts will only touch a date or hashcode at the very end. More thoughts later.

gruenich:
@westes What do you think? What steps needs automation and what better remain manual?
I am very egoistic, as I want to see this pull request reviewed and merged in any form as soon as possible. Fingers crossed that we get a flex release in a not too distant future. :heart_decoration:

westes:
> Amusingly, if we start producing nightly builds it will be less complicated to update the version numbers because the automatic parts will only touch a date or hashcode at the very end. More thoughts later.

Could you submit a pr for nightly builds then? It would be particularly helpful since we've had a number of recent changes and clearly a release is in the offing.

Mightyjo:
> > Amusingly, if we start producing nightly builds it will be less complicated to update the version numbers because the automatic parts will only touch a date or hashcode at the very end. More thoughts later.
> 
> Could you submit a pr for nightly builds then? It would be particularly helpful since we've had a number of recent changes and clearly a release is in the offing.

Yeah, I'll give it a shot.

westes:
> Yeah, I'll give it a shot.

Thank you. Just the tar ball that make distcheck would build. Make any assumptions you need to make it easier to do.

Mightyjo:
Quick update to say I'm almost there. This has been interesting. Pretty much every API changed since I wrote the milestone version.

You can see how it's working on my fork. The tarballs are attached to the action runs for Nightly. They persist for 90 days right now. I plan to attach them to tags. Still figuring out that API and the needed token permissions.

Croydon:
> I plan to attach them to tags. 

In my personal opinion, I wouldn't do that. Nightly builds that create tags or even GitHub Releases just spam everyone watching them in some way (e.g. via notifications)

Mightyjo:
Fair point. I was thinking about how to announce successful nightlies and keep their links together. Ideas welcome!

westes:
I'd suggest making a nightly release and replacing files upon successive nightly releases. (There is an action that does that if you wanna go that route.)

Croydon:
> I'd suggest making a nightly release and replacing files upon successive nightly releases. (There is an action that does that if you wanna go that route.)

Sounds good and avoids spam :D 

Alternatively, it could be a way to create some new `nightly/index.html` file (or so) with a list of the nightlies for the GitHub Page in the `gh-pages`. It would then be accessible via https://westes.github.io/flex/nightly/

Mightyjo:
That took longer than I expected.

Nightly builds are working in my fork and uploading to a release/tag called "nightlies". Please take a loot over the next couple of days to see the files build up. I'm uploading both "Source" and distro tarballs right now. I'll probably drop the "source" stuff since it's just duplicating what's in git at the named commit hash.

I have a bit more work planned to remove duplicated script chunks from the three workflow files. I'll also refactor/squash a sequence of commits from tonight in which I was chasing my tail over a token permissions SNAFU.

gruenich:
What is needed to advance this topic and get it finally merged? A rebase? More feedback? Some reviews?

westes:
@gruenich the issues are time and money. I'm working on them both.

gruenich:
@westes Have you considered enabling GitHub sponsors?

GitMensch:
As nightly builds are available now, it seems the next part is to rebase this PR and finish it.
Because that still seems to be the main open point for the next release.

PR #491, opened 2021-06-10
Issue 469; Compute length of yy_state_buf correctly

### LINKED ISSUE #469: Flex causes memory corruption and eventually causes program to crash

Mightyjo:
New PR resolves #469. Replaces PR #475 and incorporates the lessons/discussions from that thread.

Ended up mangling my Git history and needed a new PR to make sure the patches would merge cleanly.


GitMensch:
Can you please link this PR to #469 (either by adjusting the text in the PR or possibly by manually editing the attribute on the right side) to auto-close that on pull?
While I think this PR needs a review, it would definitely be nice to have the issue fixed soon, before a potential upcoming automated release.

Mightyjo:
Please review it. @smcpeak caught a bug in my first patch. I was confident in it after that, but now I've had to re-write most of it.

The changes are small, except for the test cases themselves. Those have to be large enough to potentially overflow the state buffer. The failure isn't just based on input size, though, and I haven't figured out how to generate input that triggers it reliably. So we have these garbled files based on the shape of those from the bug report because I didn't want to include someone else's code as a test.

smcpeak:
I reviewed the changes, but did not test them.  I did not find any bugs, but there are some (apparently) unintentional file mode changes, and some code that I think could be factored a bit more.

src/c99-flex.skl:

* The file's mode is changed from 644 to 755.  That looks unintentional.  Also double-check the created files; I do not know if github would show me if a new file was created with mode 755.

* The comment says "two end-of-buffer byte states", but the code (correctly) says three.  The comment could explain why 3 and not 2.

* The code that possibly reallocates `yy_state_buf` appears twice, once in `yyrestart` and again in `yy_switch_to_buffer`.  I would look to factor that into a common routine.

* Then, the code that possibly creates `yy_state_buf` in `yylex` does logically the same thing as the preceding, but works differently because the current buffer might not yet be created.  I would instead look to create `yy_state_buf` a few lines later (right before the call to `yy_load_buffer_state`), after the current buffer has been created if necessary.  Then, creation of `yy_state_buf` could be done with by the same common alloc/realloc routine.

src/cpp-flex.skl:

* Unintentional mode change?

* This also has a comment saying "two" but it should be "three".

* It's unfortunate that all this code has to be repeated for C++.  I will note that in my port I have combined the C and C++ scanners into a single engine, and highly recommend it (as a separate effort of course).

* As in the C engine, the realloc code is duplicated.

src/gen.c:

* Unintentional mode change?

* Is this file supposed to be included?  It is empty.

src/go-flex.skl:

* Unintentional mode change?

* Again, comment says "two", should be "three".

* Realloc code is duplicated.

* As in the C engine, I think the allocation code could be moved to after the main buffer is created to consolidate it with the realloc code.

* I'm skeptical of the design that has a completely different Go skeleton.  I suspect it could be combined with the C engine without too much difficulty.  But I'm not volunteering.  :)

tests/state_buf.direct.lll:

* A negative value should not be passed to `exit` as the effect is non-portable.  Instead the value should be in [0,255].

tests/state_buf_multiple.direct.dll:

* Negative passed to `exit`.

Tests generally:

* I'm surprised the tests are so complex.  In my testing, all I needed to trigger the bug was a single rule that accumulated all available text as a single token: [scan-large-string.lex](https://github.com/smcpeak/smflex/blob/smflex/test/scan-large-string.lex).  It's of course possible I have overlooked something.

Mightyjo:
> I reviewed the changes, but did not test them. I did not find any bugs, but there are some (apparently) unintentional file mode changes, and some code that I think could be factored a bit more.

Thanks! I'm not sure what's up with the mode changes. They aren't intentional; probably a side-effect of sharing my VM's filesystem with Windows.

> * The comment says "two end-of-buffer byte states", but the code (correctly) says three.  The comment could explain why 3 and not 2.

I see it saying, "plus the start state, plus the two end-of-buffer byte states." That's three, just like you explained it to me. Let me know if the comment looks different to you.

> * The code that possibly reallocates `yy_state_buf` appears twice, once in `yyrestart` and again in `yy_switch_to_buffer`.  I would look to factor that into a common routine.

Quite a few things need refactoring in that unit of code. yy_switch_to_buffer, yypop_buffer_state, yypush_buffer_state, and yyrestart all share significant stretches of code. I intend to go back to clean that whole group up at once. I'm trying to avoid refactoring it in this patch set.

> * Then, the code that possibly creates `yy_state_buf` in `yylex` does logically the same thing as the preceding, but works differently because the current buffer might not yet be created.  I would instead look to create `yy_state_buf` a few lines later (right before the call to `yy_load_buffer_state`), after the current buffer has been created if necessary.  Then, creation of `yy_state_buf` could be done with by the same common alloc/realloc routine.
>


> src/c99-flex.skl:
> src/go-flex.skl:
> 
These are basically the same skeleton with different naming cases. The "Go" skeleton is just producing C right now. Only the cpp-flex.skl is used in practice. The backend system for these is working well, but the c99 skeleton is duplicative at the moment and the Golang skeleton is really just producing C99.

> * I'm skeptical of the design that has a completely different Go skeleton.  I suspect it could be combined with the C engine without too much difficulty.  But I'm not volunteering.  :)

Golang is close, but Python, Ruby, various DSLs aren't. Having different skeletons lets us embrace new languages more easily. Even having C and C++ in the same skeleton is a beast to maintain. 

> 
> src/gen.c:
> 
> * Unintentional mode change?
> * Is this file supposed to be included?  It is empty.
> 

This is not supposed to be empty. Have to look into this. 

> 
> tests/state_buf.direct.lll:
> 
> * A negative value should not be passed to `exit` as the effect is non-portable.  Instead the value should be in [0,255].
> 
> tests/state_buf_multiple.direct.dll:
> 
> * Negative passed to `exit`.
> 

Rgr. Forgot about this change to the test harness.

> Tests generally:
> 
> * I'm surprised the tests are so complex.  In my testing, all I needed to trigger the bug was a single rule that accumulated all available text as a single token: [scan-large-string.lex](https://github.com/smcpeak/smflex/blob/smflex/test/scan-large-string.lex).  It's of course possible I have overlooked something.

I wrote these tests before your earlier notes on clarifying whether the state buffer size is specified in elements or bytes. Since I was creating HUGE state buffers as a result of that confusion I might have needed this test case that generates a ton of extra states every time it encounters a brace. I'll try again with something simpler now that the state buffer is the expected, smaller size.

Mightyjo:
Answer on the "src/gen.c: empty file" thing. That's how GitHub surfaces a patch that only changes a file's mode. You'll see that gen.c is no longer affected by this patch set, but tests/ruleset.am had the wrong permissions and is showing that "empty file" status now. Nothing to worry about, just UX weirdness.

Mightyjo:
> * Then, the code that possibly creates `yy_state_buf` in `yylex` does logically the same thing as the preceding, but works differently because the current buffer might not yet be created.  I would instead look to create `yy_state_buf` a few lines later (right before the call to `yy_load_buffer_state`), after the current buffer has been created if necessary.  Then, creation of `yy_state_buf` could be done with by the same common alloc/realloc routine.

I'm losing my mind a little of the chunk of code you're talking about here. I've written a better initializer half a dozen times and Git keeps losing it during rebase. I would have missed it being lost again without you, so thanks! 

My init code was still present in c99-flex.skl this time. Ported it back to cpp-flex.skl and made your suggested improvement to the control flow in all three skeletons.

smcpeak:
> I see it saying, "plus the start state, plus the two end-of-buffer byte states." That's three, just like you explained it to me.

My mistake!  You're right, I misread it.  In fact I misread it multiple times because I was surprised at the discrepancy.  Reading comprehension fail.

> I intend to go back to clean that whole group up at once. I'm trying to avoid refactoring it in this patch set.

I agree with avoiding refactoring existing code at the same time as fixing a bug.  But here, in each skeleton, there are three new blocks of code that are similar or identical.  What I'm suggesting is to create (in each skeleton) one new function, something like (untested!):

```
static void yy_possibly_realloc_state_buf(yyscan_t yyscanner)
{
	if ( yyscanner->yy_state_buf_max < (yy_current_buffer(yyscanner)->yy_buf_size + YY_STATE_BUF_EXTRA_SPACE) ) {
		size_t new_size = yy_current_buffer(yyscanner)->yy_buf_size + YY_STATE_BUF_EXTRA_SPACE;
		yy_state_type *new_state_buf = (yy_state_type*) yyrealloc( yyscanner->yy_state_buf, (new_size * sizeof(yy_state_type)), yyscanner );

		if ( new_state_buf == NULL ) {
			yypanic( "out of dynamic memory in yylex()", yyscanner );
		}
		else {
			yyscanner->yy_state_buf = new_state_buf;
			yyscanner->yy_state_buf_max = new_size;
		}
	}
}
```

then in each of the three places that currently add a block of code like that (for example, right after `yy_load_buffer_state( yyscanner );`), just insert a call to `yy_possibly_realloc_state_buf(yyscanner)`.

> Answer on the "src/gen.c: empty file" thing. That's how GitHub surfaces a patch that only changes a file's mode.

Got it, thanks.  I wondered if that might be the case.

Mightyjo:
You're right. I'll work on refactoring the common logic.

PR #492, opened 2021-06-10
Codeql errors

Mightyjo:
Fixed errors I was seeing in the new CodeQL action in master. This produces reports on flex itself, now.

NB: We're looking pretty good according to these reports.


Mightyjo:
And you can see it works because it passed its own checks! I should have expected that, but it still gave me a laugh.

westes:
This is now merged in to master. Thank you!

PR #493, opened 2021-06-12
Build amendments

jannick0:
@westes @Mightyjo Please review. Locally all tests succeed. However, `git2cl` appears to be not too happy about how the commit messages are formatted, i.e. `ChangeLog` does not look great.  But I think the commits are OK.


Explorer09 (commented):
Please don't do this. `BUILT_SOURCES` is a bad way to fix build dependency problems where explicit dependency rules in makefile do it better. The main problem of `BUILT_SOURCES` is that it's executed in all `make` targets (except cleaning targets) including libfl-only build. You'll break more with `BUILT_SOURCES` than it fixes.

Mightyjo (commented):
We try to support a variety of older systems, so we try to support the oldest version of the build dependencies that we can. 

Why do we need to move to Autoconf 2.69 from 2.60?

Mightyjo (approved):
RE: [c217658](https://github.com/westes/flex/pull/493/commits/c217658e78fceade9633177a1c409529b44dc9e8)
The define-then-subst idiom makes the variable available to checks within the configure script. Doesn't look like we're using pkgconfigdir anywhere else right now so this is fine.

Mightyjo (changes requested):
Re: [dd68f07](https://github.com/westes/flex/pull/493/commits/dd68f0770dc57995c396bd09f3723cc2c8732a5b)
Don't use the phrase "BUILT_SOURCES" (or variations) for this change. Its meaning is special in Automake - referring to built files dependency tracking can't handle. "FLEX_RULESETSOUCES" or something like that would be fine and won't collide with Automake variables.

I'm not seeing any warnings about the $(FLEX) dependencies on the suffix rules. Recommend leaving those in place.

Mightyjo (changes requested):
Re: [d13b1d5](https://github.com/westes/flex/pull/493/commits/d13b1d5c968f4c18a3a7107fed71d15292ff6b62)
Use $(SKELINCLUDES) variable in CLEANFILES instead of $(BUILT_SOURCES). The way BUILT_SOURCES is used here is already suspect since Automake's dependency tracking should be able to build those files as needed from the rules. The files that require them might just not be linked up correctly.

Mightyjo (changes requested):
Re: [4356a4a](https://github.com/westes/flex/pull/493/commits/4356a4ad3f4557bdf8cd5a6aa6a173b4a38ffc74)
Don't do this. BUILT_SOURCES should only be used with Automake is unable to tack dependencies any other way. It was correct before.

Also, don't remove bootstrapping support in this changeset.

Mightyjo (changes requested):
Re: [6ae0982](https://github.com/westes/flex/pull/493/commits/6ae0982eb772e1a5b2116e4ded3f76e7c634bd81)
I don't think this does what you think. libflex.a gets built once and used for both stage1flex and flex. That defeats the purpose of boostrapping. I ran into this, too. It's frustrating from an aesthetic standpoint, but it's practically no time compared to building the test suite.

Mightyjo (commented):
Re: [cab48e1](https://github.com/westes/flex/pull/493/commits/cab48e161e2068c3160e854284b89979ef68eaa2)
This is a good reason to bump up a dependency version.

Mightyjo (changes requested):
Re: [c8b9ab1](https://github.com/westes/flex/pull/493/commits/c8b9ab112fb09388b0a0d02964cde6d30e8e7f3e)
Related to an earlier commit. This isn't needed if you don't change the way suffix rules depend on $(FLEX).

Mightyjo (changes requested):
Re: [21492a2](https://github.com/westes/flex/pull/493/commits/21492a2628b6e1c3d9a754d93fd5391517c19d54)
If the branch regarding re-used CCLs is dead, why don't we remove it?

Explorer09:
For your info, we used to have a problem with the `BUILT_SOURCES` automake variable. See #144. Now I always prefer explicit dependency tracking instead of that variable.

westes:
Please split this up into multiple, small, atomic pull requests. Because, some of the things you want to do are clearly good ideas and easy to do. Because some of the things you want to do are good ideas but should be implemented differently. Because some of the things you want to need to be discussed and sorted out to see if that's what flex should do.

Explorer09 (commented):
Is @jannick0 still interested in updating this branch?
Since @westes had merged some of the pull requests I made, allow me to state some updates to this branch:
* 78b92f5ecfd206f91ec137d9335a14c1ef2ff0aa "build(flex): remove non-portable GNU MAKE extension" -> Superceded by a8b07bac95ae09160a79854c73a67297b467b827
* b3dae9fe726d7c743f9786d4c77180944fcb03b2 "build(flex): add less verbose mode for generated files" -> The `$(SED)` variable is no longer needed as the sedding was replaced with 2e5f4bc9d8f2d39ad162b9b50882e20cf4154ead ("build: Generate deterministic scan.c and stage1scan.c.") I'll leave it to you on the `$(AM_V_GEN)` and `$(M4)` variable parts, as I cannot verify whether your patch would work correctly.
* d13b1d5c968f4c18a3a7107fed71d15292ff6b62 "build(flex): simplify target CLEAN" -> Superceded by bab5ce6457be396b2f91ea8b04a14faa750780db (build: let "make mostlyclean" delete "*-flex.h".)
* 4356a4ad3f4557bdf8cd5a6aa6a173b4a38ffc74 "build(flex): add bison c-files to BUILT_SOURCES for parallelized builds" -> Please drop this one. As mentioned in the previous comments, we are avoiding BUILT_SOURCES with a reason. It's better for makefiles to explicitly state file dependencies than grouping files that should be built before others as a vague hint. Also, don't delete `scan.c` as part of `make clean` target! You would force users to re-extract the tarball to get the file back. As for `parse.c` dependency, it is not needed. Automake recognizes `parse.y` in `flex_SOURCES` and thus is `parse.c` is covered implicitly.

westes:
@Explorer09 If you want to update this pr (or make a new one) to account for the more recent updates, please feel free to do so.

Explorer09:
@westes I personally don't know how to work with the branch of @jannick0 as his vision of improving flex is different from mine. And if it's me to improve flex's build system, I would try making a different goal such as "non-recursive make". And thus I have no interest in taking this one over.

PR #495, opened 2021-06-22
fix compile warning originating in cpp skeleton

GitMensch:
`gcount()` is of type `streamsize` which has a system-dependent size with at least `int`. Because of the `int max_size` limit no bigger value will be returned, the explicit cast fixes warnings from compilers/static analyzers.

As an alternative `LexerInput` may be changed to have a return and parameter of type `streamsize`, but this would need more changes and likely create more warnings in other places.

This issue originates from https://github.com/lexxmark/winflexbison/issues/73 where more details can be found (this is not a MSVC only issue).

### DIFF
--- a/src/cpp-flex.skl
+++ b/src/cpp-flex.skl
@@ -2390,7 +2390,7 @@ int yyFlexLexer::LexerInput( char* buf, int max_size )
 	if ( yyin.bad() ) {
 		return -1;
 	} else {
-		return yyin.gcount();
+		return (int)yyin.gcount();
 	}
 #endif
 }


Mightyjo (approved):
Looks good. Your reasoning makes sense. I double-check that streamsize is defined as a "signed, integral type", so with our limit to int max_size this seems right. 

'git clean -xf' ... 'make distcheck' from a clean maintainer tree build cleanly.

GitMensch:
@Mightyjo Thanks for the review, so it is up to @westes to merge.

Side question @westes: maybe there's someone that could be added as a developer to this repo and could work on handling the PRs (currently 39 open) / issues (>100)?

Issue #367, opened 2018-06-20
inclusion of WinFlex changes (drop need for fork/wait and posix conversion function htonl/htons)


GitMensch:
https://github.com/lexxmark/winflexbison/tree/master/flex/src includes some changes for non-GCC (obviously especially WIN32) compilers.

What do you think about integrating them (after cleanup)?

A good starting example seems to be filter.c (and flexdef.h) to use a different signature of `struct filter` and `filter_apply_chain ()`. These changes could be integrated if one or both functions `fork()`/`wait()` are not available (at least with 2.6.4 there is a check in configure for them but neither a configure time failure nor a `#ifdef HAVE_FORK` in the code).

Similar could be done for tables.c and `htonl` / `htons` (which seem to have no check during configure).

So... was a discussion about this before?

GitMensch:
Note: if this is agreed upon I'd add some PR with getting similar changes in that are needed to compile and run with OrangeC and VS (as it seems to be the main target of the WinFlexBison changes).

Explorer09:
@GitMensch I have discovered the non-portability of htonl / htons, but I'm not sure how WinFlexBison solves it. (htonl and htons are used in skeleton, so adding a check of them in configure is useless. We need other workaround for the case.) Can you give a better pointer?

Regarding the others, I can't comment at the moment.

Explorer09:
@GitMensch 
Maybe I was wrong about htons/htonl pair, but their inverse, `ntohs` and `ntohl`, have to be included in the skeleton as it is part of the decoder of the flex table file format.

I believe making htons/htonl macro defines would cause more portability problem than it solves, since (I think) some compilers would recognize htons/htonl functions and optimize them into inline assembly without generating function calls. Repeating what a compiler would do in a header is a bad idea.

I think with portability to Windows, including <winsock.h> would be the way to go. But for even more exotic platforms, I don't know if we would need this naive fallback:

```c
short htons(short val) {
    unsigned char arr[sizeof(short)] = {0};
    int i;
    for (i = sizeof(short) - 1; i >= 0; i--) {
        arr[i] = (unsigned char)(val);
        val >>= CHAR_BIT;
    }
    return (*(short *)arr);
}
```

(Yes, you get it. This is the most portable htons implementation within ANSI C constraints.)

westes:
If we're going to provide an implementation of a function, we should use the libobj mechanism.

[...]

PR #514, opened 2022-01-10
improve compatibility for _MSC_VER

GitMensch:
copy from lexxmark/winflexbison@2a7586a by @yjh-styx - partial work on #367

### DIFF
--- a/src/tables.c
+++ b/src/tables.c
@@ -36,6 +36,11 @@
 #include "flexdef.h"
 #include "tables.h"
 
+#ifdef _MSC_VER
+#define htonl(n) _byteswap_ulong(n)
+#define htons(n) _byteswap_ushort(n)
+#endif
+
 /** Convert size_t to t_flag.
  *  @param n in {1,2,4}
  *  @return YYTD_DATA*. 


Explorer09:
I think this missed a check that MSVC compiles code only for little-endian machines. Otherwise the semantic would be wrong.
`htonl` and `htons` should be no-op when compiled for big-endian and become byte swap functions for little-endian.

Explorer09:
Update: I think it would be good enough to cite this in the code comments
https://learn.microsoft.com/en-us/cpp/standard-library/bit-enum?view=msvc-170
"All native scalar types are little-endian for the platforms that Microsoft Visual C++ targets"

GitMensch:
So you suggest to make those macros a no-op, right?

Explorer09:
@GitMensch No. The byte order does need to be swapped for little endian. What I meant is add a comment that MSVC only supports little endian platforms and you have the macros defined the way you did them.

```c
#ifdef _MSC_VER
/* MSVC only supports little endian. 
(https://learn.microsoft.com/en-us/cpp/standard-library/bit-enum?view=msvc-170) */
#define htonl(n) _byteswap_ulong(n)
#define htons(n) _byteswap_ushort(n)
#endif
```

GitMensch:
I see, thank you for the clarification.
This PR is nearly 14 months old, but if someone with pull access wants this change I'd do it of course.

Croydon:
@Explorer09 you can use GitHub's code suggestion feature, then @GitMensch can apply the changes with a few clicks and you get credits :)

Explorer09:
> you can use GitHub's code suggestion feature, then @GitMensch can apply the changes with a few clicks and you get credits :)

I was expecting that @GitMensch adjust the formatting of my suggested change instead of applying the changes as is.
Note that `_byteswap_ulong` and `_byteswap_ushort` depend on `<stdlib.h>` in MSVC's build environment, so it's not just a simple function (macro) replacement.

westes:
If someone wants to emplement @xplorer09's suggestion (with whatever else MS stuff needs to be done inside the ifdef), I would merge it.

PR #523, opened 2022-05-11
GitHub Actions (Replacing TravisCI)

Mightyjo:
These are the changes from #480 that convert automated builds from Travis CI to Github Actions. A couple of new commits update the scripts to run on Ubuntu 20.04.

I'll rebase and resubmit the release automation change set once this set is settled.


Mightyjo:
The result of the automated build (with distcheck) and CodeQL analysis are above. I'm awfully glad they succeeded :)

Croydon (approved):
LGTM

westes (approved):

GitMensch:
That is really a good step forward. It raises two issues though:

* So far you can't see from the outside if you use a 2.6.4 or a newer ci snapshot. This is because configure.ac still has the version 2.6.4 - can this be adjusted to be 2.6.5-dev, please? This is mostly a question @westes (of course the build could, before "make distcheck" also adjust the version and append something like "ci-$(date -Iminutes)".
* Can you please add artifacts (something to download)? For builds passing this would be flex*.tar.*.

Mightyjo:
> * So far you can't see from the outside if you use a 2.6.4 or a newer ci snapshot. This is because configure.ac still has the version 2.6.4 - can this be adjusted to be 2.6.5-dev, please? This is mostly a question @westes (of course the build could, before "make distcheck" also adjust the version and append something like "ci-$(date -Iminutes)".
> * Can you please add artifacts (something to download)? For builds passing this would be flex*.tar.*.

Good points! Take a look at #480 from which this was split out. This workflow is mainly to support code review of new PRs. The ones in #480 are close to what you're asking for. I'm looking for discussion there about the use of semantic-release, etc.

PR #528, opened 2022-05-26
fix some tiny doc errors.

xiaohanyu:
It seems that the release number is `2.6.4` instead of `4.6.2`, and, there are some grammar issue in the doc.


Piyaphartoi:
xiaohanyu:fix

westes (approved):

PR #530, opened 2022-05-31
Add error messages to unreachable ends of switch statements

### LINKED ISSUE #103: still some warnings with -Wswitch-default

ggujjula:
This PR fixes #103 by adding error messages (either `yypanic()` or `YY_FATAL_ERROR`) as a default case in switch statements that check against the return value of `yy_get_next_buffer()`.

Since `yy_get_next_buffer()` is defined to return only one of three values, and all these switch statements have cases for all of these values, it would be an error if execution fell through to the default case. Additionally, adding these statements helps future-proof the codebase if `yy_get_next_buffer()` is ever modified.

### DIFF
--- a/src/c99-flex.skl
+++ b/src/c99-flex.skl
@@ -1214,6 +1214,8 @@ int yyinput (yyscan_t yyscanner)
 			case EOB_ACT_CONTINUE_SCAN:
 				yyscanner->yy_c_buf_p = yyscanner->yytext_ptr + offset;
 				break;
+			default:
+				yypanic("unexpected return value from yy_get_next_buffer()", yyscanner);
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


Explorer09:
Can you assure the cases are truly unreachable? I would prefer an assertion error rather than a fatal error whose code would even present in the release build.

PR #542, opened 2022-09-27
INSTALL file

westes:
document build/installation process


Mightyjo:
Looks good. Only one note:

INSTALL:30 DESDIR should be DESTDIR.

I forgot you already had all those dependency details in README. That was handy.

PR #544, opened 2022-10-04
Call `input` rather than `yyinput` in src/scan.l:

burke:
This allows the stage1 build to succeed even when using a currently-released (2.6.4) version of Flex.

Open to suggestions of how this could be done better if calling the legacy name is undersirable.

### DIFF
--- a/src/scan.l
+++ b/src/scan.l
@@ -673,7 +673,10 @@ void context_member(char *, const char *);
                    */
                 &&  (cclval = ccllookup( nmstr )) != 0 )
 				{
-				if ( yyinput() != ']' )
+				/* Intentionally using the old API rather than `yyinput`, so that
+				 * we can bootstrap using flex 2.6.4 and older.
+				 */
+				if ( input() != ']' )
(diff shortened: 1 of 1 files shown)


Mightyjo:
What's your build environment where this was failing? It works as is in our CI/CD build.

burke:
I'm on macOS, and observed this behaviour with both `flex 2.6.4 Apple(flex-34)` at `/usr/bin/flex` and `flex 2.6.4` from Homebrew.

Mightyjo:
Copy that, and thanks! Haven't been able to test on macOS for a while. (Although I think I saw it was back in GH Actions, so maybe soon.)

Explorer09:
I don't see it's a good idea if we call the old API without some kind of version check. By the way, it's the commit dad680611b8373cac2e5dc60e7552535bded79c2 that deprecated `input()` due to its name collision in C++ mode.

Explorer09:
Why close this PR? Does the issue no longer exist or is it something else?

PR #554, opened 2023-02-20
Fix Issue #539 crash on Apple M1 by casting 0 to (char *) explicitly

### LINKED ISSUE #539: Compile failed on MacOS 12.4 (arm chip)
clang 13 or gcc 12 both compile error 
```
./cpp-flex.h:9:69: error: expected expression
  "m4_define([[M4_PROPERTY_TRACE_LINE_REGEXP]], [[^#line ([0-9]+) ",\"(.*)\"]])
                                                                    ^
./cpp-flex.h:9:70: warning: missing terminating '"' character [-Winvalid-pp-token]
  "m4_define([[M4_PROPERTY_TRACE_LINE_REGEXP]], [[^#line ([0-9]+) ",\"(.*)\"]])

```

It look like m4 process error.

My m4 version is 

image

and generate `cpp-flex.h`

image

As I'm not familiar with m4, why it generate the ",\" ?

biergaizi:
Currently, when the NULL-terminated variadic function `filter_create_ext()` is invoked, the value `0` is passed as the last argument to act as a terminator. However, this is an integer value, which is incompatible with the pointer data type expected by `filter_create_ext()`.

This is undefined behavior in C, correct operation is not guaranteed. In fact, it causes flex to crash on Apple M1 when GCC is used - the loop is not terminated when it should, instead, it keeps running, corrupting the argument list for invoking m4. As a result, it creates the following error:

> flex: fatal internal error, exec of gm4 failed

This commit fixes the problem by explicitly casting the value `0` to the correct pointer type `(char *)`. According to Rich Felker, the primary author of musl libc, [type casting from 0 is technically safer than using the value `NULL`](https://ewontfix.com/11/).

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -295,7 +295,7 @@ void initialize_output_filters(void)
 	if ( !(m4 = getenv("M4"))) {
 		m4 = M4;
 	}
-	filter_create_ext(output_chain, m4, "-P", 0);
+	filter_create_ext(output_chain, m4, "-P", (char *) 0);
 	filter_create_int(output_chain, filter_fix_linedirs, NULL);
 
 	/* For debugging, only run the requested number of filters. */


biergaizi:
Also see https://github.com/biergaizi/flex/commit/cce2df853386d5b5b60445b1204dcca08e9f259e for backport to v2.6.4.

PR #555, opened 2023-03-01
remove unused DO_COMPARISON

GitMensch:
that was used in older versions of the script, but now seems to be only set to zero here and not used anywhere


westes:
Please rebase and resolve conflicts; I'd like to get this merged in.

GitMensch:
done

PR #566, opened 2023-06-03
Update README

6NoFantasy9:


westes:
Could you fix the other typo on that line as well (and then squash your commits down)?

PR #582, opened 2023-09-06
Hash-pin GitHub Actions, keep them updated with dependabot

### LINKED ISSUE #581: Keep GitHub Actions used in workflows hash-pinned
Hey, it's Pedro (see #562) and I'm back with a new security suggestion!

When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

These hashes can be automatically updated by dependabot. Whenever new versions Actions are released, you'll receive a single PR updating all of their hashes and version comments at once (see [this example](https://github.com/pnacht/libarchive/pull/1)).

I'll send a PR pinning the Actions along with this issue.

pnacht:
Fixes #581.

This PR hash-pins GitHub Actions to protect the project from potential supply-chain threats.

The Actions will be updated by dependabot. dependabot will send a single PR updating all Actions with new versions at once.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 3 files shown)


westes:
Thanks, it's now in!

PR #595, opened 2023-10-02
add apt-get update to ci

vmeshche:
`apt-get install` failed with `E: Failed to fetch`. Looks like some old deb staff in cache and we need run `apt-get update` before setapping env.


Mightyjo:
Looks like github fixed this by updating their runner images.

westes:
In general, this approach makes builds less reproduceable, not more. This also makes it harder to tell what the underlying OS contains. If we're concerned that particular versions of packages should or should not be present, then we should simply specify what we need rather than adding a step that makes the process take longer and be less determined.

PR #599, opened 2023-10-19
Skip linenum increment for indented code blocks

taniwha:
An indented code block would cause the subsequent all subsequent code block line numbers (including its own) to be out by one.

For example, the following would cause the line number for the `BEGIN` code block to have the line number for the `warning` code block. This would also affect the code in the final section after the second `%%`.

    %%
                yyflexdebug = 0;
                grab_frame = GRAB_FRAME;
                grab_other = GRAB_OTHER;
                grab_write = GRAB_WRITE;
                yylval->pointer = 0;// ensure pointer vals are null

    "/*"        { BEGIN (COMMENT); }
    <COMMENT>"/*" { warning (0, "nested /* in comment"); }

### DIFF
--- a/src/scan.l
+++ b/src/scan.l
@@ -95,7 +95,7 @@ extern const char *escaped_qstart, *escaped_qend;
 #define START_CODEBLOCK(x) do { \
     /* Emit the needed line directive... */\
     if (indented_code == false) { \
-        linenum++; \
+        if (!x) linenum++; \
         line_directive_out(NULL, infilename, linenum);	\
     } \
     add_action(M4QSTART); \
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -80,7 +80,7 @@ lexcompat*
 !lexcompat.rules
 !lexcompat.txt
 lineno*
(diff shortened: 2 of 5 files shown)


westes:
This should have a test case to detect regression in the future. Is that someting you're able to add?

taniwha:
Sure, I'll look into that.

taniwha:
Done. I covered as many places as I could think of, and I verified that it does catch the incorrect lines before the patch.

westes:
What do you think about having the lexer be generated from something that e.g. grep -n 's the line numbers so that changing the source is less prone to break the test?

taniwha:
Oh, hmm, I hadn't thought of that. Interesting idea. I'll see what I can come up with, but generation certainly makes sense.

taniwha:
I came up with something that works nicely so long as the two (`@LINE@` and `__LINE__`) are on the same line.

taniwha:
@westes are there any further issues?

westes:
This looks great; thank you!

PR #608, opened 2023-11-16
Allow white space in yymore and yyreject checks

taniwha:
AC_PROG_LEX puts a space between yymore and its ()s causing the yymore check to fail and thus break configure.

### DIFF
--- a/src/scan.l
+++ b/src/scan.l
@@ -927,11 +927,11 @@ nmstr[yyleng - 2 - end_is_ws] = '\0';  /* chop trailing brace */
 			ACTION_ECHO;
 			CHECK_REJECT(yytext);
         }
-		"yyreject()" {
+		"yyreject"{OPTWS}"("{OPTWS}")" {
 			add_action("]""]M4_HOOK_REJECT[""[");
 			CHECK_YYREJECT(yytext);
         }
-		"yymore()" {
+		"yymore"{OPTWS}"("{OPTWS}")" {
 			yymore_used = true;
 			if (ctrl.rewrite)
 				context_call(yytext);


westes:
I'm confused on where/when configure is broken. I just rebuilt configure and ran it from my local copy of the git repo. It completed successfully.

d125q:
It is broken in the sense that AC_PROG_LEX does not detect flex built from HEAD (8453b08 at the time of writing).

AC_PROG_LEX generates the following program:

```c
%{
#ifdef __cplusplus
extern "C"
#endif
int yywrap(void);
%}
%%
a { ECHO; }
b { REJECT; }
c { yymore (); }
d { yyless (1); }
e { /* IRIX 6.5 flex 2.5.4 underquotes its yyless argument.  */
#ifdef __cplusplus
    yyless ((yyinput () != 0));
#else
    yyless ((input () != 0));
#endif
  }
f { unput (yytext[0]); }
. { BEGIN INITIAL; }
%%
#ifdef YYTEXT_POINTER
extern char *yytext;
#endif
int
yywrap (void)
{
  return 1;
}
int
main (void)
{
  return ! yylex ();
}
```

As you can see, the parentheses are preceded by a space.  This causes flex to fail with:

```txt
conftest.l:10:3: error: use of undeclared identifier 'yymore_used_but_not_detected'
{ yymore (); }
  ^
conftest.c:614:18: note: expanded from macro 'yymore'
#define yymore() yymore_used_but_not_detected
                 ^
lex.yy.c:1704:25: warning: equality comparison with extraneous parentheses [-Wparentheses-equality]
                if ( (yy_buffer_stack == NULL) ) {
                      ~~~~~~~~~~~~~~~~^~~~~~~
lex.yy.c:1704:25: note: remove extraneous parentheses around the comparison to silence this warning
                if ( (yy_buffer_stack == NULL) ) {
                     ~                ^      ~
lex.yy.c:1704:25: note: use '=' to turn this equality comparison into an assignment
                if ( (yy_buffer_stack == NULL) ) {
                                      ^~
                                      =
1 warning and 1 error generated.
configure:14088: $? = 1
```

For the record, I was attempting to build [bedeb7dc](https://git.savannah.gnu.org/cgit/wget.git/tree/configure.ac?id=bedeb7dc27bf3407c22b9d278f08fa91d7e80146#n269) of wget.

westes:
What are the exact steps I need to follow to reproduce the problem?

d125q:
Create a `configure.ac` file:

```
AC_INIT
AC_PROG_LEX([noyywrap])
AS_IF([test "x$LEX" != xflex], [AC_MSG_ERROR([flex not found])])
AC_OUTPUT
```

Run:

```
$ autoreconf -ifv
$ ./configure
```

Result:

- Works with flex tag v2.6.4 as well as the patch in this PR.
- Fails with HEAD (8453b08 at the time of writing).

Mightyjo:
Prior to 2.6.4, we didn't include the parentheses in the yymore check at all. I think returning to that construction is the best solution.

This patch will also work, but it prefers C-like function calls. Since I'm slowly working on other output language backends I'm biased toward solutions that will work more generally.

Thanks for pointing this out!

westes:
Researching, commit 191d706 is where the particular change was made for yymore.

Mightyjo:
Sitting with this for a few months, I'm happy with @d125q's solution. I was concerned that using {OPTWS} in the rules would lead to line length problems like those in #620, but we already use this construction in scan.l without trouble so I'm withdraw that concern.

If we ever support an action language that doesn't put parentheses around its argument/parameter lists, I expect we'll be able to search for this thread.

westes:
I'm working up a test case for this. It's subtler to sort out that I'd like, so I want to be sure we're able to catch it from happening again.

westes:
Thanks for this; it's now on master.

PR #625, opened 2024-02-14
Changes and Edits in the README File

Lin-Selina:
-Edited the README file documentation
-No code edited
-Kept the original content
-Added in more structure and additional information to help users who would like to use this project

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,37 @@
 # Flex: the Fast Lexical Analyzer Generator
 [![Build Status](https://github.com/westes/flex/actions/workflows/build.yml/badge.svg)](https://github.com/westes/flex/actions/workflows/build.yml)
 
-flex is a tool for generating scanners: programs which recognize lexical patterns in text.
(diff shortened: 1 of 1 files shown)


westes:
according to github, the TOC should show up automagically.

westes:
I'm not sure that I think of thise document as a web-first document, so the links to files aren't really a thing I've worried about.

westes:
It's not clear how the explanation of scanners and flex helps people who know they need to use flex.

Mightyjo:
Retracting my statement that this deletes some original content. With one exception: "This is flex, the fast lexical analyzer generator." It's iconic and should remain intact.  I see where the explanation of the repository structure ended up, now.

I don't think linking to top level files and directories from README.md is useful. Particularly since GitHub added the repo navigator bar to the left side of the screen.

I don't love this explanation of scanner applications. It conflates the roles of scanners and parsers in its examples. That's already confusing enough for new folks without us throwing murky water in the mix. I'd rather say nothing or point newcomers to a good book.

westes:
As noted in the comments, this pull request has multiple problems. This pull request does not address any current needs of the flex project. As such, I am closing this pull request.

PR #639, opened 2024-03-30
build: Use portable alternative to $'\r' in mkskel.sh

Explorer09:
(This is a follow-up to #353 . I can't reopen the old PR, so I file another one to address my point.)

Replace the `$'\r'` quoting in commit
c674b5faa3419032a6c41cc150488866f3b3cff9 with a portable alternative `` cr=`printf '\r'` ``

Although `$' '` (dollar-single-quote) quoting is proposed for inclusion in POSIX, we have not seen it in a published standard yet, so it is best to avoid it for now.


westes:
Thank you!

PR #640, opened 2024-03-30
build: Makefile improvements with skeleton sources

Explorer09:
1. Correct skeleton headers dependencies
2. Don't use BUILT_SOURCES for skeleton files 
3. Let "make mostlyclean" delete "*-flex.h"


westes:
Thanks for this! It's on master.

PR #649, opened 2024-05-08
build: Update AX_PROG_CC_FOR_BUILD to fix BUILD_EXEEXT problems

Explorer09:
The BUILD_EXEEXT and BUILD_OBJEXT variables in `configure` was broken.
The ax_prog_cc_for_build.m4 from autoconf-archive upstream contains some fixes to BUILD_EXEEXT and BUILD_OBJEXT, but they were broken again with Autoconf 2.70.

In this PR, I updated the ax_prog_cc_for_build.m4 module while added my own fixes to make BUILD_EXEEXT and BUILD_OBJEXT work in Autoconf 2.70.

(I also submitted the patches upstream, but they are not merged yet:
https://savannah.gnu.org/patch/index.php?10452
https://github.com/autoconf-archive/autoconf-archive/pull/296)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -78,6 +78,15 @@ AC_PROG_LN_S
 AC_PROG_AWK
 AC_PROG_INSTALL
 
+AS_IF([test "x${BUILD_OBJEXT-}" = x],
+  [AS_IF([test "$cross_compiling" = no],
+    [BUILD_EXEEXT="$EXEEXT"
+    BUILD_OBJEXT="$OBJEXT"],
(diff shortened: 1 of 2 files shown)


westes:
What is the problem that you're fixing though?

Explorer09:
@westes I was trying to rewrite the bootstrap mechanism in the makefiles so that the bootstrap can be more portable -- even when the object file extensions are different (BUILD_EXEEXT != EXEEXT). It's not in a hurry to merge this one, as Flex's makefiles didn't use BUILD_EXEEXT at the moment, but I plan to use it in the rewrite.

PR #675, opened 2024-10-03
Making flex re-entrant and releasing all acquired resources at the end

mingodad:
This is an experiment that replace all globals by a dynamic allocated structure to hold the state that is passed everywhere and also releasing all acquired resources (memory/file handles) at the end.

The main drive behind this changes is to create something like https://mingodad.github.io/parsertl-playground/playground/ using this fork and https://github.com/mingodad/lalr-parser-test/tree/main/byacc (also with globals replaced) or doing a similar work with bison.

If anything here is useful to the original project feel free to grab it.

**All tests are passing.**

If someone want to try build this branch be aware that the build will fail when trying to parse the generated `parse.c` at that point you need to change to the `src` folder and issue this command `make fix-parse` then again `make` and `make check` to see hopefully all tests passing.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -228,4 +228,7 @@ indent: $(top_srcdir)/.indent.pro
 		echo $$f FAILED to indent; \
 	done;
 
-.PHONY: indent
+fix-parse: parse.c
+	sed -i -E -e 's/(yy[^,]+,[^,]+), gv\)/\1, yyscanner, gv)/' parse.c
+
+.PHONY: indent fix-parse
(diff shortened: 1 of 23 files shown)


Mightyjo:
Flex itself is a compiler-like tool. It isn't meant to be re-entrant. The code it generates is re-entrant, though.

We collected the state into structures a few years ago. That was good. I'm not seeing the value in allocating that state dynamically and changing every prototype in the code base.

westes:
Please let us know when your build process runs successfully.

PR #679, opened 2024-10-16
Create Scanner

akanocaleb:


Mightyjo:
What are you improving or changing? You can link to issues and PRs by number. (This one is #679, for example.)

If you're adding an example, it should live in the examples/ directory. Your scanner is quite similar to examples/manual/numbers.lex which you may find interesting.

Finally, Flex input files typically have the suffix ".l" or ".lex".

westes:
Closing as per @Mightyjo's comments.

PR #702, opened 2025-03-20
build(tests): add missing parser->scanner dependencies

rossburton:
With sufficient parallelism and `make --shuffle` to reorder builds the test suite occasionally fails to build with errors like this:

 tests/bison_yylval_parser.y:36:10: fatal error: bison_yylval_scanner.h: No such file or directory
    36 | #include "bison_yylval_scanner.h"

Solve this by adding the missing dependencies to the *_parser.o targets on the _scanner.h files.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -334,6 +334,7 @@ lineno_generated.c: lineno_generated.l
 
 bison_nr_main.$(OBJEXT): bison_nr_parser.h bison_nr_scanner.h
 bison_nr_scanner.$(OBJEXT): bison_nr_parser.h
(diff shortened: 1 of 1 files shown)


Mightyjo:
Thanks for pointing out the --shuffle option! I hadn't heard about that yet.

How much parallelism have you needed to trigger this? I've run up to -j8 without seeing it.

rossburton:
-j64 triggers it on my machine, but this machine also has 128 cores...

Mightyjo:
Nice.

Thanks for finding this!

westes:
Thanks for this; a good catch.

PR #704, opened 2025-05-16
fix: noline for %top directives

### LINKED ISSUE #463: #line directive is output even when %option noline is used
When adding this to a lex file:

```
%option noline
%top{
    #include <cstdint>
}
```
or, equivalently, using -L on the commant line, the output file will always include #line for the %top section.

jjmaestro:
Fixes #463 and #268

When using the --noline flag or the %option noline, flex was failing to suppress the %top line directives.

There was an if-guard missing in the code that adds the line directives to the top_buf. This commit fixes it.

### DIFF
--- a/src/scan.l
+++ b/src/scan.l
@@ -177,7 +177,8 @@ void context_member(char *, const char *);
 		snprintf(trampoline, sizeof(trampoline),
 			 "M4_HOOK_TRACE_LINE_FORMAT(%d, [[%s]])",
 			 linenum, infilename?infilename:"<stdin>");
-                buf_strappend(&top_buf, trampoline);
+                if (ctrl.gen_line_dirs)
+                    buf_strappend(&top_buf, trampoline);
                 brace_depth = 1;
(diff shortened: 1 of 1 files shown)


jjmaestro:
@westes could you approve the workflows? I've manually run the checks with `make checks` and all passed (I had 3 skipped, bison tests).

Also, I see there hasn't been a release in a while... could the fix be considered for a `2.6.5` release? 🙏  Thanks!!

westes (commented):
Thanks for this. I approved the workflows which have now completed successfully. Merged as well.

This will be in the 2.6.5 release. There are a couple things I need to get done to make that happen. We're much closer to that then we were.

jjmaestro:
@westes awesome, thanks!!

jjmaestro:
@westes quick ping about this! I think there's no more open issues for `2.6.5` except the "Automate the release" milestone, I think. Do you think it's ready for release? :) Thanks!!

westes:
Yes, it's in my queue.

PR #717, opened 2026-02-08
test\!: UNDO Turn off tests for fake Go backend

Mightyjo:
This prevents the mock Go skeleton currently on the mainline from causing test errors. It should be reverted when a real Go skeleton and backend are implemented.

Split from #710


westes:
This one turned out to be second.

Mightyjo:
Second is ready.

PR #731, opened 2026-04-29
skel: validate yytables_fload header buffer and bound table dimensions

### LINKED ISSUE #730: Three bugs in `yytables_fload` (binary tables-file loader)
Found these in `src/cpp-flex.skl` while writing a libFuzzer harness against the runtime tables loader. All three trigger from `yytables_fload(fp, scanner)` (the public API a consumer calls when `%option tables-file=` is used). Reproduced against the unmodified loader in master @ 13c8018; same code is present in 2.6.4 unchanged.

Two heap-OOB reads and one integer-driven multi-GB allocation. ASAN traces below; PoC bytes are at the bottom of this issue.

## 1. heap-OOB read in `strlen` (yytbl_hdr_read)

`src/cpp-flex.skl:4010`

```c
th->th_name = th->th_version + strlen (th->th_version) + 1;
```

`th_version` was just `fread`'d (line 4002) into a `yyalloc(bytes)` buffer (line 3997), where `bytes = th_hsize - 14` is set at line 3996. Nothing guarantees those bytes contain a NUL. The header sanity check at line 3990 only requires `hsize >= 16`, so the smallest payload is 2 bytes. With those 2 bytes set to `'A','A'`, `strlen` walks past the heap allocation.

ASAN with a 16-byte input file:

```
==58898==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000000d2
READ of size 3 at 0x6020000000d2 thread T0
    #0 strlen
    #1 cl_tables_fload  lex.cl_.c:2701
    #2 main             poc_strlen_oob.c:101

0x6020000000d2 is located 0 bytes after 2-byte region
allocated by thread T0 here:
    #0 malloc
    #1 cl_tables_fload  lex.cl_.c:2701
```

(The `lex.cl_.c:2701` frame is the public `cl_tables_fload` after Apple clang inlined `yytbl_hdr_read` and `yytbl_fload`. Internal frames separate at `-O0 -fno-inline`.)

## 2. heap-OOB read in `strcmp` (yytbl_fload)

`src/cpp-flex.skl:4278`

```c
if (strcmp(th.th_name, key) != 0) {
```

Different root cause from #1. Even when the version section IS NUL-terminated, `th_name` is set unconditionally to `th_version + strlen + 1`. If the version's NUL lands at the LAST byte of the buffer, `th_name` points just past the heap region, and `strcmp` walks past it. PoC: 16-byte input where `hsize=16`, payload is `'A','\0'`. ASAN reports `READ of size 1` past a 2-byte region, attributed to `cl_tables_fload`.

The fix for #1 and #2 is the same shape: validate that `th_version` and `th_name` each find a NUL inside the buffer. Roughly:

```c
size_t vlen = strnlen(th_version, bytes);
if (vlen >= bytes - 1) { yyfree(th_version,...); return -1; }
size_t nlen = strnlen(th_version + vlen + 1, bytes - vlen - 1);
if (nlen >= bytes - vlen - 1) { yyfree(th_version,...); return -1; }
th_name = th_version + vlen + 1;
```

## 3. integer-driven multi-GB allocation (yytbl_data_load)

`src/cpp-flex.skl:4070-4072`

```c
if ((td.td_flags & YYTD_STRUCT)) {
    bytes = sizeof(struct yy_trans_info) * td.td_lolen * (td.td_hilen ? td.td_hilen : 1);
} else {
    bytes = td.td_lolen * (td.td_hilen ? td.td_hilen : 1) * dmap->dm_sz;
}
... yyalloc(bytes, ...);
```

`td_lolen` and `td_hilen` are `flex_uint32_t` read straight from the file, no validation. A 50-byte tables file with `td_lolen = 0xF1238000` makes the loader call `malloc(8091271168)` (7.54 GB) for one record. On any consumer with realistic memory limits (containers, embedded, mobile) this is fatal: either the OOM-killer takes the process or `malloc` returns NULL and `YY_FATAL_ERROR("out of dynamic memory in yytbl_data_load()")` calls `exit()`.

There's already a precedent fix for the build-time tool's allocators in commit `9c54eb6 build: detect overflow for [re]allocate_array`. The runtime loader was never given the same guard. A two-line clamp before the multiplication would mirror that commit.

## Reproduction

PoCs are standalone C `main()` programs that link the unmodified `lex.cl_.c` from `flex --tables-file=cl.tab sample.l`:

```
clang -g -O1 -fno-omit-frame-pointer \
      -fsanitize=address,undefined -fno-sanitize-recover=undefined \
      poc_strlen_oob.c lex.cl_.c -o poc_strlen_oob
ASAN_OPTIONS=detect_leaks=0:halt_on_error=1 ./poc_strlen_oob
```

All three reproduce in <10ms through the public `cl_tables_fload` (≡ `yytables_fload`). PoC files attached. Affects only scanners generated with `%option tables-file=` or `--tables-file=`.

## Suggested fixes

1. and 2.: replace `strlen` with bounded `strnlen` and verify both NULs fall inside the buffer (snippet above).
3. clamp `td_lolen` and `td_hilen` to a sane upper bound before the multiplication, in the same spirit as `9c54eb6`. Something like `if (td_lolen > YYTBL_MAX_DIM || td_hilen > YYTBL_MAX_DIM) return -1;` with `YYTBL_MAX_DIM` set to whatever real scanners actually need (a few million is more than enough; 32k DFA-states is the documented limit).

Happy to send a PR covering all three if you'd like.

Found via `OSS-Fuzz`-shaped libFuzzer harness, expanding what the upstream OSS-Fuzz `projects/flex/` integration covers (the upstream `flex-patch.diff` neuters the m4 fork, which incidentally also keeps the loader off the fuzzed code path; that's why none of these have surfaced from continuous fuzzing).

MarkLee131:
Fix https://github.com/westes/flex/issues/730: The binary tables-file loader trusts attacker-controlled bytes from the file in three places.  
  1. yytbl_hdr_read at cpp-flex.skl:4010 did th->th_name = th->th_version + strlen(th->th_version) + 1; immediately after fread'ing th_hsize - 14 bytes into th_version. If those bytes contain no NUL, strlen walks past the heap region. Fix: use strnlen() and require both the version and the name string to terminate with NULs inside the buffer; return -1 otherwise.

  2. yytbl_fload at cpp-flex.skl:4278 then did if (strcmp(th.th_name, key) != 0) even when th_name == th_version + bytes, where the buffer has no room for the name section.  Fixed by the same validation in (1): th_name is only set when a name NUL is found inside the buffer.

  3. yytbl_data_load at cpp-flex.skl:4090-4092 computes bytes = sizeof(yy_trans_info) * td_lolen * td_hilen (or the equivalent non-STRUCT form) with td_lolen and td_hilen read directly from the file as flex_uint32_t.  A 50-byte tables file with td_lolen near UINT32_MAX makes the loader call yyalloc(8 GB).  Fix: reject td_lolen or td_hilen above YYTBL_MAX_DIM (1<<24) before the multiplication, mirroring 9c54eb6 which added the analogous guard to the build-time tool's allocators.

Verified with three minimum-size PoCs (16, 16, 50 bytes) that the unpatched loader trips ASAN on each, and the patched loader returns -1 cleanly.  A valid tables file generated by `flex --tables-file=` still loads as before (regression check passed).

### DIFF
--- a/src/cpp-flex.skl
+++ b/src/cpp-flex.skl
@@ -4007,7 +4007,27 @@ static int yytbl_hdr_read YYFARGS2(struct yytbl_hdr *, th, struct yytbl_reader *
 	} else {
 		rd->bread += (flex_uint32_t) bytes;
 	}
-	th->th_name = th->th_version + strlen (th->th_version) + 1;
+	/* Validate that the buffer contains TWO NUL-terminated strings within
+	 * its bounds (a version string followed by a name string).  Without
+	 * these checks, a header whose payload is missing either NUL would
+	 * cause the strlen() below, and the strcmp() in yytbl_fload, to read
+	 * past the heap allocation. */
+	{
+		size_t vlen, nlen;
+		vlen = strnlen (th->th_version, bytes);
+		if (vlen >= bytes - 1) {
+			yyfree (th->th_version M4_YY_CALL_LAST_ARG);
+			th->th_version = NULL;
+			return -1;
+		}
+		nlen = strnlen (th->th_version + vlen + 1, bytes - vlen - 1);
+		if (nlen >= bytes - vlen - 1) {
+			yyfree (th->th_version M4_YY_CALL_LAST_ARG);
+			th->th_version = NULL;
+			return -1;
+		}
+		th->th_name = th->th_version + vlen + 1;
+	}
 	return 0;
 }
 
@@ -4045,6 +4065,19 @@ static int yytbl_data_load YYFARGS2(struct yytbl_dmap *, dmap, struct yytbl_read
 	    || yytbl_read16 (&td.td_flags, rd) != 0
 	    || yytbl_read32 (&td.td_hilen, rd) != 0
 	    || yytbl_read32 (&td.td_lolen, rd) != 0) {
+		return -1;
+	}
+	/* Bound the dimensions before any size_t multiplication.  Without
+	 * this, a malicious tables file with td_lolen and td_hilen set to
+	 * arbitrary uint32 values drives the bytes computation below into
+	 * a multi-GB allocation request.  YYTBL_MAX_DIM is well above what
+	 * any real scanner produces (the documented NFA-state limit is
+	 * around 32k; tables larger than a few million elements would not
+	 * fit any realistic memory budget). */
+#ifndef YYTBL_MAX_DIM
+#define YYTBL_MAX_DIM (1u << 24)
+#endif
+	if (td.td_lolen > YYTBL_MAX_DIM || td.td_hilen > YYTBL_MAX_DIM) {
 		return -1;
 	}
 	/* Lookup the map for the transition table so we have it in case we need it


jpaulin:
Does it break anything with a relative good -coverage test set? Can I help here?

PR #742, opened 2026-06-27
Fix vscode warnings for includes in main

HattonLe:
Using VisualCode, a few minor warnings come up in main.c, so added a couple of missing includes to resolve those.
Also removed an include for "parse.h" which isn't in the repo anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ flex-*/
 libtool
 m4/
 stamp-*
+/.vscode/
+
--- a/src/main.c
+++ b/src/main.c
@@ -35,10 +35,10 @@
 
 
 #include "flexdef.h"
+#include "config.h"
 #include "version.h"
 #include "options.h"
 #include "tables.h"
-#include "parse.h"
 
 static char flex_version[] = FLEX_VERSION;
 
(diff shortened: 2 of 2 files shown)


Mightyjo:
Welcome! 

Flex uses a bunch of generated code from automake, bison, and flex itself. 
- `parse.h` is generated by bison from `parse.y`. Don't remove the include in `main.c` if you want compilation to work.
- `config.h` is generated from `config_for_build.h.in` by make after we bootstrap. It's included in `main.c` through `flexdef.h` which includes additional logic for grabbing the right header depending on the bootstrapping stage.
- `locale.h` is also included through `flexdef.h` where we check additional condition flags and set up gettext.

All that is to say, VSCode will show you fewer include errors after you configure the source directory. It will show even fewer errors after you build the first time. I use it too and I've just gotten used to ignoring certain warnings as IDE quirks.

westes:
Please take the time to understand a codebase before trying to make changes to it.

HattonLe:
Well "parse.h" isn't actually used in "main.c". It compiled fine for me.

HattonLe:
As there isn't a repo discussion page....

How to I get printf() to work in the code.... it seems to be crapping out during compiling with a "stage1scan.c:1:1: error: unknown type name ..." error. etc...

westes:
> As there isn't a repo discussion page....
> 
> How to I get printf() to work in the code.... it seems to be crapping out during compiling with a "stage1scan.c:1:1: error: unknown type name ..." error. etc...

A closed pull request is not the place to ask for basic programming assistance.

Please consult your textbooks, your teachers, etc. for how to perform basic programming tasks.

PR #743, opened 2026-06-28
New flex C++ header file generation process.

HattonLe:
Removed dependency on use of external file "FlexLexer.h".
Flex now generates "FlexLexer.h" using its internal resource copy of that header file, now split up into two parts for base and derived class definitions.
Derived class names now generated at source without needing the #define junk in the header.
Further refinement of class definitions can now be made by editing the templates in "src/FlexLexerB.txt" and "src/FlexLexerD.txt".
Make uses xxd to compress the templates for use as an embedded flex resource.
Tested on Ubuntu build only.


Mightyjo (changes requested):
Re: [bed9bf2](https://github.com/westes/flex/pull/743/commits/bed9bf25f49ab1c205044b7debe6253681b8bc21) We rejected these changes in a previous PR. Please remove them from this and future PRs.

Mightyjo (approved):
Re: [4f4f5dc](https://github.com/westes/flex/pull/743/commits/4fef5dc7e7e21a209dc2345d7a53d9dce41e55fa) Goodbye debug line. Why not delete it and the comment above it?

Mightyjo (changes requested):
Re: [44d92c4](https://github.com/westes/flex/commit/44d92c4e139d26b2e4abc3791da10f16376decb9) Squash this into [52ea572](https://github.com/westes/flex/commit/52ea5727ce3f3c16365dfdb68dabdc4569762ef1), please.

Mightyjo (changes requested):
Re: [52ea572](https://github.com/westes/flex/commit/52ea5727ce3f3c16365dfdb68dabdc4569762ef1)

- Remove the changes to the `.po` files, please. We avoid committing this unless we've actually changed the translations. I need to add a note about this in CONTRIBUTING.md. My fault.
- Please don't add a dependency on `xxd`. I'd ask that you not obscure the contents of the `flexTemplate*.h` files at all. You can turn the sources into C string headers the same way we do with the `.skl` files.
- Could you use the existing `src/FlexLexer.h` as the source of truth for this class generation? It's the same one you've seen installed at `/usr/lib/FlexLexer.h` (maybe a different path depending on your distro). It's always been our header, it just gets installed rather than emitted.
- NB: You weren't seeing output from `printf()` because Flex reopens `stdout` to point at the output `FILE *`. (It goes through the filter chain first, see `filter.c`.) Try including `flexdef.h` and using the `lerr()` method. It prints tombstones and continues. We also have `flexerror` if you want to print and terminate. The other variants do different amounts of cleanup before termination.

Mightyjo (commented):
Before you get too far with the implementation, did you know we already produce a C++ class when you set the language to c++? (With the `-+` command line option or `%option c++` in the input `.l` file.)

HattonLe:
Re: [bed9bf2] I forgot to remove the previous local branch before I started this change so its already been dealt with (I hope)

Re: [4f4f5dc] I eventually figured something was redirecting stdout, so already removed the printf() which I left in by mistake. After all thats what your're for... spotting mistakes. I just noticed the goodbye.. haha.

[52ea572] I though git was ignoring the PO files - git status said none of them were listed as staged when I pushed so I don't know why they went up. Before pushing is there something to do like a dist-clean or whatever to stop junk going to remote?

If there's a better method than xxd let me know. It was the easiest to get the requirement done at the time but I can factor it out if there is a standard way to acheive the same result.

My defacto "/usr/lib/FlexLexer.h" isn't the same as what was in this repo (I haven't done any "install" for flex as I don't want to break my existing platform, and I don't know where "/usr/lib/FlexLexer.h" came from as this is the first time I'm using flex in anger since lex wasn't cutting it). The difference in the header file caused my application to fail when running flex from the repo instead of the platform lex. Hence if the flex executable created that file it would never be out of sync due to other installed lexes\flexes.. and that makes it easier to use\maintain as well.

lerr() tombstones... yes I found that out... will switch to using flexerror.

Yes I know it already generates a cpp file.... and I'm not redoing all of that. But that header file is a dogs dinner, and using all those undefine\defines for multiple classes is a real bodge. No wonder that might be causing issues like breaking inheritance for some people. I want to get to the point where it will be a very simple task to refactor the existing cpp generating code to use the auto generated headers.

Updated my repo fork to bug fix \ and to update the generated class file content for smoother 'refactoring' with the existing flex generated .cpp files. Also added auto generation of a "FlexLexer.cpp" for base class \ one off C implementation code (such as yyalloc(), yyfree() and yyrealloc()).

Its taken a lot of the boilerplate C\C++ junk out of FlexLexer.h, and removed some junk out of the generated cpp file too, which makes understanding it and maintenance a lot easier.
As POC (proof of concept) I'm almost down to just renaming by hand all the cpp file occurences of "yyFlexLexer" to get it to compile with the new files.

Mightyjo:
> Re: [[bed9bf2](https://github.com/westes/flex/commit/bed9bf25f49ab1c205044b7debe6253681b8bc21)] I forgot to remove the previous local branch before I started this change so its already been dealt with (I hope)
> 

Copy that. No worries.

> [[52ea572](https://github.com/westes/flex/commit/52ea5727ce3f3c16365dfdb68dabdc4569762ef1)] I though git was ignoring the PO files - git status said none of them were listed as staged when I pushed so I don't know why they went up. Before pushing is there something to do like a dist-clean or whatever to stop junk going to remote?
> 

Weird. I don't know gettext very well, unfortunately. 

After committing my changes, I'll usually do `git clean -xdf` to throw away any side effects. That's also the preferred starting point for `make check` and `make distcheck`.

> If there's a better method than xxd let me know. It was the easiest to get the requirement done at the time but I can factor it out if there is a standard way to acheive the same result.
> 

Look at `src/mkskel.sh`. That turns the `.skl` emission templates into header files included by `src/skeletons.c`. The `src/cpp-flex.skl` is the default for generating both C and C++ right now. We've been working on separating those into their own skeletons for a while. Long story short, it touches most of the source code so it's taking a while.

> My defacto "/usr/lib/FlexLexer.h" isn't the same as what was in this repo (I haven't done any "install" for flex as I don't want to break my existing platform, and I don't know where "/usr/lib/FlexLexer.h" came from as this is the first time I'm using flex in anger since lex wasn't cutting it). The difference in the header file caused my application to fail when running flex from the repo instead of the platform lex. Hence if the flex executable created that file it would never be out of sync due to other installed lexes\flexes.. and that makes it easier to use\maintain as well.
> 

Thought that might be what's up. Something like that brought me to the project years ago. 

> lerr() tombstones... yes I found that out... will switch to using flexerror.
> 
> Yes I know it already generates a cpp file.... and I'm not redoing all of that. But that header file is a dogs dinner, and using all those undefine\defines for multiple classes is a real bodge. No wonder that might be causing issues like breaking inheritance for some people. I want to get to the point where it will be a very simple task to refactor the existing cpp generating code to use the auto generated headers.
>

Copy that. It's not an idiomatic C++ solution. The multiple inclusion technique isn't unique to Flex, though. We even have regression tests to ensure it's working correctly, see `tests/cxx_multiple_scanners.cc`. What's the inheritance problem you've seen? I don't recall seeing it posted to the issues here.
 
> Updated my repo fork to bug fix \ and to update the generated class file content for smoother 'refactoring' with the existing flex generated .cpp files. Also added auto generation of a "FlexLexer.cpp" for base class \ one off C implementation code (such as yyalloc(), yyfree() and yyrealloc()).
> 
> Its taken a lot of the boilerplate C\C++ junk out of FlexLexer.h, and removed some junk out of the generated cpp file too, which makes understanding it and maintenance a lot easier. As POC (proof of concept) I'm almost down to just renaming by hand all the cpp file occurences of "yyFlexLexer" to get it to compile with the new files.

Okies, I'll leave you to it. Holler if you need help with new tests, distcheck, the GitHub automation, etc.

HattonLe:
Thanks for the feedback,useful stuff to know. I rarely use git so sometimes I end up in a mess if I get deviated from my cheatsheet of useful git commands. Added "git clean -xdf" to my cheatsheet :o)

Is is possible to delete that first 'fix' of mine  [bed9bf2](https://github.com/westes/flex/pull/743/commits/bed9bf25f49ab1c205044b7debe6253681b8bc21) ? I don't know github workflow enough for how to get rid of that.

Was looking at when flex finishes writing out the cpp file - I want to run something after that file has been written...
If I breakpoint on main ~line 685 (FLEX_EXIT) the majority of the cpp file content is present (looking at it in realtime via Notepadqq), but it doesn't get flushed until I run flex to end of execution.
Is there a method I can call to finish the cpp file processing, i.e. flush it to disk before FLEX_EXIT?
Updated my repo fork just now from midnight code session with cleaned up source from feedback and improved error handling.

westes:
The current pr says it's over 5700 lines changed.

Please split this up into atomic smaller parts.

And ensure you have test coverage so we can ensure that flex doesn't regress.

HattonLe:
I estimate 1533 lines so far. I think git went screwball the other day and uploaded a load of PO file stuff which wasn't staged by me cos I checked beforehand with a git status (I assume thats what caused the 5700 lines).

I'm designing this so it doesn't break existing functionality and can be used in parallel with old file generation.
I'll have a consistent experimental push done as soon as I have figured out one problem, namely...

Oh and 1 problem I am seeing everry so often.... occasionally when adding in a simple line of perfectly valid C code, e.g. "wait(0)"; it seems to kill the build process completely with some kind of syntax error. As far as I can tell build seems to be doing an execution or scan scan.l? of the source code? It is using flex itself to auto generate stuff for the build or something?

HattonLe:
I reached a sensible point in my repo to do a checkpoint of it. Feel free to try that version out from my repo.

- No changes to original flex functionality, my code only runs if the --c++ option was specified on the command line.
- My output files generated with filenames prefixed with "DEV-" to distinguish them as new and experimental.
- Also output is a auto-editied version of the original flex .cpp output to show what a fully finished cpp file looks like.

"DEV-Flexer.*" contained the base class and any one-off C definitions\declarations. No dependancies on existing flex header files.

For an example flex run using "--yyclass=abc", "DEV-abc.cpp" would be the auto-editied version of the original flex "abc.cpp" output file.

If  you were to run diff or meld between the two output .cpp files, e.g.  "DEV-abc.cpp" and "abc.cpp" you would be able to see what changes would be needed to the original flex cpp output.

With the current way flex runs its impossible to get access to the entire flex cpp output file contents. That why the auto-editied file is missing the bottom 5% of the flex cpp output file. (its a m4 execv child process issue I haven't yet figured out how to work around).

The auto-editor (and its "DEV-*.cpp" output file) are temporary demo artifacts for work in progress visualisation only.
The DEV prefix would eventually be removed once parallel running\validation in the wild has been done and after the original flex cpp generator is re-worked so it matches the auto-edited output file.

I need to get my head around this skeletons stuff and the app dataflow - this app is a lot more complex that I thought it would be.

westes:
Please resubmit this when you have gotten the change set down to a reasonable number of lines.

